PR #222, opened 2023-06-14
data integration: Pass timestamp field to the source

aditya-nambiar:


bothra90 (commented):
Looks good! Requesting one minor change to field name

fennel/gen/connector_pb2.pyi
```diff
@@ -579,6 +580,7 @@ class Source(google.protobuf.message.Message):
     cursor: builtins.str
     @property
     def lateness(self) -> google.protobuf.duration_pb2.Duration: ...
+    ts_sort_column: builtins.str
```
bothra90, line 583: Can we call this "timestamp_field" instead? That's more inline with terminology in other places. Also, the usage of this for sorting is only specific to some sources.

bothra90 (approved):

PR #223, opened 2023-06-16
[DEV-1405] add definition apis to the py client

showbufire:


aditya-nambiar:
Looks good to me, but since the client is now public and this would be a page which folks refer to for documentation, can you add doc comments describing each API. The parameters and what to expect from each. 

Feel free to edit the description of other API's if they dont meet the bar.

fennel/client/client.py
```diff
@@ -128,6 +128,84 @@ def log(
             response = self._post_json("{}/log".format(V1_API), req)
         return response
```
aditya-nambiar, line 130: Looks good to me, but since the client is now public and this would be a page which folks refer to for documentation, can you add doc comments describing each API. The parameters and what to expect from each.

Feel free to edit the description of other API's if they dont meet the bar.

aditya-nambiar (changes requested):
Looks good to me, but since the client is now public and this would be a page which folks refer to for documentation, can you add doc comments describing each API. The parameters and what to expect from each.

Feel free to edit the description of other API's if they dont meet the bar.

aditya-nambiar (approved):

PR #224, opened 2023-06-16
[DEV-1406] add lookup api to py client

showbufire:


fennel/client/client.py
```diff
@@ -301,6 +301,22 @@ def extract_historical_features(
         else:
             return pd.Series(response.json())
 
+    def lookup(
+        self, dataset_name: str, keys: List[Dict[str, Any]], fields: List[str]
+    ) -> tuple[Union[pd.DataFrame, pd.Series], pd.Series]:
+        req = {
```
aditya-nambiar, line 307: Same comment as other PR

fennel/client/client.py
```diff
@@ -301,6 +301,22 @@ def extract_historical_features(
...
+            "keys": keys,
```
aditya-nambiar, line 308: Now that we have this function, can we modify tests in client_tests/test_dataset.py to replace Dataset.lookup to client.lookup ?

fennel/client/client.py
```diff
@@ -301,6 +301,22 @@ def extract_historical_features(
...
+            "fields": fields,
+        }
```
aditya-nambiar, line 310: This api only looks up latest ? why not also have asof semantics?

fennel/client/client.py
showbufire: done

fennel/client/client.py
showbufire: Took a brief look, I don't think it's that straightforward. The `client.lookup` requires the running of the our def server.

fennel/client/client.py
showbufire: That's a good idea. The current rest server doesn't have the asof supported, probably because it's not there yet when we started the metadata API project.

aditya-nambiar (approved):

PR #235, opened 2023-07-07
lambda function: Fix lambda function extraction for python 3.11

aditya-nambiar:
We earlier used byte matching to find the lambda text, but now use a more reliable source of information i.e the offsets in the lambda ast node


bothra90 (approved):
Please add a description to PR as to why this fixes the issue

PR #237, opened 2023-07-11
aggregate: Enable count unique for the client

aditya-nambiar:


docs/pages/api-reference/aggregations.md
```diff
@@ -15,7 +15,11 @@ Besides these common arguments, here is the rest of the API reference for all th
 
 ### 1. Count
 
-Count computes a rolling count for each group key across a window. Takes no additional arguments besides `window` and `into_field`. It returns 0 by default. Its output type is always `int`.&#x20;
+Count computes a rolling count for each group key across a window.  It returns 0 by default. Its output type is always `int`.&#x20;
+The count aggregate also takes an optional argument `unique` which is a boolean. If set to true, it counts the number of unique values in the given window.&#x20;
+The field over which the count is computed is specified by the `of` parameter of type `str`.&#x20;
+Since the unique count uses a HyperLogLog algorithm to estimate the number of unique values and hence is an approximate count.&#x20;
+Hence please set the `approx` parameter to true while using this feature.&#x20;
```
nikhilgarg28, line 22: This should be rewritten to have a different narrative. You can say something to the effect of "approx is another argument which when turned on makes the answer approximate but allows Fennel to be more efficient with state storage. As of right now, Fennel only supports approx unique count so if you set unique=True, you'd also need to set approx=True.

fennel/client_tests/test_dataset.py
```diff
@@ -953,7 +964,15 @@ def filter_positive_ratings(cls, rating: Dataset):
...
         )
         return filter2.groupby("movie").aggregate(
-            [Count(window=Window("forever"), into_field=str(cls.cnt_rating))],
+            [
+                Count(window=Window("forever"), into_field=str(cls.cnt_rating)),
+                Count(
+                    window=Window("forever"),
+                    into_field=str(cls.unique_ratings),
+                    of="rating",
+                    unique=True,
```
nikhilgarg28, line 973: This didn't set approx=True. Should this be an error?

fennel/client_tests/test_dataset.py
```diff
@@ -862,6 +864,12 @@ def pipeline_aggregate(cls, activity: Dataset):
                 Count(
                     window=Window("forever"), into_field=str(cls.total_ratings)
                 ),
+                Count(
+                    window=Window("4d"),
+                    into_field="unique_ratings_4d",
+                    of="rating",
+                    unique=True,
```
nikhilgarg28, line 871: This didn't set approx=True, should this be an error?

fennel/test_lib/execute_aggregation.py
```diff
@@ -60,6 +60,27 @@ def get_val(self):
...
+
+    def del_val_from_state(self, val):
+        self.counter[val] -= 1
+        if self.counter[val] == 0:
+            del self.counter[val]
+        self.count = len(self.counter)
+        return self.count
+
+    def get_val(self):
+        return self.count
```
nikhilgarg28, line 81: To ensure that people expect 'approx' to work, will it make sense to arbitrarily add/remove a random amount upto 1% or so? We just need to be careful that for the same counter, same answer (even if it's approx) is returned on each call.

nikhilgarg28 (changes requested):

fennel/test_lib/execute_aggregation.py
aditya-nambiar: I belive that is unnecessary complexity. We are making them write approx=True to make them aware.

nikhilgarg28 (approved):

PR #245, opened 2023-07-21
Fix group id bug on kafka source

showbufire:


showbufire:
discussed, we'll get rid out of the group id, and generate id on the server side

PR #246, opened 2023-07-23
schema: Add struct type to our list of supported types

aditya-nambiar:


fennel/datasets/datasets.py
```diff
@@ -594,8 +599,14 @@ def dataset(
         file_name = ""
 
     def _create_lookup_function(
-        cls_name: str, key_fields: List[str]
+        cls_name: str, key_fields: List[str], struct_types: Dict[str, Any] = {}
```
nikhilgarg28, line 602: It's usually a bad idea to use mutable data (e.g. lists, dicts) as default arguments. Better to make it None and convert it to empty dict when None.

nikhilgarg28 (changes requested):
Requesting a small change. Rest everything looks phenomenal. Thank you!

nikhilgarg28 (approved):

PR #250, opened 2023-07-29
DEV-1608 client side integration for standard deviation

zsid60:


fennel/client_tests/test_dataset.py
```diff
@@ -863,6 +879,11 @@ def pipeline_aggregate(cls, activity: Dataset):
                 Count(
                     window=Window("forever"), into_field=str(cls.total_ratings)
                 ),
+                Stddev(
+                    window=Window("3d"),
+                    of="rating",
+                    into_field=str(cls.std_rating_3d),
+                ),
             ]
```
aditya-nambiar, line 887: Can we add a 7d window too and ensure that the sd deviations are indeed different

fennel/client_tests/test_dataset.py
```diff
@@ -402,6 +404,11 @@ def pipeline_aggregate(cls, activity: Dataset):
                     into_field=str(cls.max_ratings),
                     default=0.0,
                 ),
+                Stddev(
```
aditya-nambiar, line 407: Can we add some tests in test_invalid_dataset.py where you try to break Stddev and ensures it breaks, like taking std dev of string field etc

fennel/lib/aggregate/aggregate.py
```diff
@@ -159,6 +159,22 @@ def signature(self):
         return f"lastk_{self.of}_{self.window.signature()}"
 
 
+class Stddev(AggregateType):
+    of: str
+
+    def to_proto(self):
```
aditya-nambiar, line 165: Could you add a test for to_proto in test_dataset.py ( it essentially tests if the proto generated is correct )

aditya-nambiar:
Can you also bump up the version in pyproject.toml add a changelog entry and add stddev to documentation

aditya-nambiar (approved):
add some tests for setting the default value in std dev

PR #252, opened 2023-07-30
docs: add documentation for the new 'first' operator

bothra90:
image

### DIFF
--- a/docs/pages/api-reference/operators.md
+++ b/docs/pages/api-reference/operators.md
@@ -101,6 +101,19 @@ The `aggregate` operator has the following parameters:
 
 <pre snippet="api-reference/operators_ref#aggregate"></pre>
 
+### Groupby/ First
+
+Fennel allows you to groupby and get the first entry (by timestamp) in each group using the `first` operator.
+The api requires you to chain the `groupby` operator with the `first` operator.
+
+The `groupby` operator takes the following parameters:
+
+1. `*args: str` - positional arguments that specify the list of fields to group by.
+
+The `first` operator does not take any parameters.
+
+The returned dataset's fields are the same as the input dataset, with the grouping fields as the keys.
+
 ### Drop
 
 Fennel allows you to drop columns from a dataset using the `drop` operator.
--- a/docs/pages/api-reference/operators.md
+++ b/docs/pages/api-reference/operators.md
@@ -104,15 +104,17 @@ The `aggregate` operator has the following parameters:
 ### Groupby/ First
 
 Fennel allows you to groupby and get the first entry (by timestamp) in each group using the `first` operator.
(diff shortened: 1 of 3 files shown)


docs/pages/api-reference/operators.md
```diff
@@ -101,6 +101,19 @@ The `aggregate` operator has the following parameters:
 
 <pre snippet="api-reference/operators_ref#aggregate"></pre>
 
+### Groupby/ First
```
nikhilgarg28, line 104: Please add a code snippet example?

docs/pages/api-reference/operators.md
```diff
@@ -101,6 +101,19 @@ The `aggregate` operator has the following parameters:
...
+
+Fennel allows you to groupby and get the first entry (by timestamp) in each group using the `first` operator.
+The api requires you to chain the `groupby` operator with the `first` operator.
```
nikhilgarg28, line 107: 'API requires you to...' feels a bit weird from language POV. Instead, maybe simplify and say '`...first operator must always be preceeded by a groupby (else an error is raised)...`

docs/pages/api-reference/operators.md
```diff
@@ -101,6 +101,19 @@ The `aggregate` operator has the following parameters:
...
+
+The `groupby` operator takes the following parameters:
+
+1. `*args: str` - positional arguments that specify the list of fields to group by.
```
nikhilgarg28, line 111: Not sure if this is needed here? Having this in two places makes it harder to keep the documentation upto date as the groupby operator evolves.

nikhilgarg28 (changes requested):

docs/pages/api-reference/operators.md
aditya-nambiar: Fyi we dont have a seperate groupby operator documentation, it is combined with aggregate

docs/pages/api-reference/operators.md
bothra90: Missed that. Will add something

bothra90: Added an example. I am not able to visualize the docs on my local environment. If your setup is working, can you preview and make sure it looks good?

docs/pages/api-reference/operators.md
bothra90: changed the wording. Let me know what you think now

docs/pages/api-reference/operators.md
bothra90: +1 to what Nikhil is saying (about having documentaiton for groupby duplicated), but also `groupby in itself is not a full operator (e.g. we can't end a pipeline at `groupby`) which makes it challenging. Any ideas for this @nikhilgarg28 ?

nikhilgarg28 (approved):

PR #258, opened 2023-08-10
Fixes for client after columnar change

zsid60:
There are a couple todo's still here for me to test, but we should merge this in so that changes on the server can be consumed correctly

* Test against the examples/ directory again
* look at the test_log_dataset integration test
* unit test to_columnar_json

### DIFF
--- a/fennel/utils.py
+++ b/fennel/utils.py
@@ -9,6 +9,7 @@ import json
 import sys
 import textwrap
 
+from pandas import DataFrame
 from typing import Any
 from typing import cast, Callable, Dict, List, Tuple, Union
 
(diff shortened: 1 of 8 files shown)


fennel/utils.py
```diff
@@ -207,3 +208,21 @@ def propogate_fennel_attributes(src: Any, dest: Any):
     for k, v in src.__dict__.items():
         if k.startswith("__fennel") and k.endswith("__"):
             setattr(dest, k, v)
+
+
+def to_columnar_json(df: DataFrame) -> str:
```
mohitreddy1996, line 213: maybe few unit tests for this?

mohitreddy1996:
> There are a couple todo's still here for me to test, but we should merge this in so that changes on the server can be consumed correctly

Maybe list down the TODOs too

PR #259, opened 2023-08-10
fast follow fixes for converting to columnar json in the client

zsid60:


fennel/client_tests/test_dataset.py
```diff
@@ -146,8 +146,8 @@ def test_log_to_dataset(self, client):
         assert response.status_code == requests.codes.BAD_REQUEST
         if client.is_integration_client():
             assert (
-                response.json()["error"]
-                == """error: input parse error: expected Int, but got String("32")"""
```
zsid60, line 150: this is now allowed (as of 6/10/23). In the server we try to parse to int and if it fails then send the error

aditya-nambiar (approved):

PR #260, opened 2023-08-10
DEV-1696 re allow output feature list to be a featureset

zsid60:


bothra90 (changes requested):
We don't want to allow input list to have featuresets since that can break existing applications using the REST API. Only the output list should be allowed to have featuresets

zsid60:
@bothra90: fixed to only change for output

bothra90 (approved):

zsid60:
TODO: update changelog / version

PR #262, opened 2023-08-12
S3 connector: Expose option to set presorted=True

aditya-nambiar:


fennel/lib/to_proto/to_proto.py
```diff
@@ -579,6 +581,7 @@ def _s3_to_ext_table_proto(
     path_prefix: Optional[str],
     delimiter: str,
     format: str,
+    pre_sorted: bool,
```
nikhilgarg28, line 584: Let's rename this to `timesorted` or something that's one word + easier to understand.

nikhilgarg28 (changes requested):

PR #265, opened 2023-08-14
DEV-1699 unit test for df -> columnar dict utility function

zsid60:
Unit test follow up to #259

`pd.to_json()` doesn't have a builtin way of the exact format we wanted: `dict[col] -> [values]`
`pd.to_dict` has this option, but `to_dict` handles `NaNs`, `Nones`, and `timestamps` differently from `to_json`.
Therefore we provide a custom function for converting dicts to columnar json. This PR provides unit tests for that function. 

I also added a couple typo fixes in the docs pages that I found as I was reading through the docs

### DIFF
--- a/fennel/test_utils.py
+++ b/fennel/test_utils.py
@@ -1,5 +1,7 @@
 from fennel.utils import fhash, to_columnar_json
 
+import pandas as pd
+
 
 def test_fhash_Callable():
     def f(x: int, y: int) -> int:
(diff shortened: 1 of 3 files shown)


fennel/test_utils.py
```diff
@@ -32,3 +37,57 @@ def f(x: int, y: int) -> int:
...
+    now_ms = int(now.timestamp() * 1000)
+    expected = {
+        "num": [1.0, None],
+        "word": ["abc", "def"],
+        "bool": [None, False],
+        "math": [3.14, 2.71828],
+        "time": [now_ms, now_ms],
+    }
+    # str encodes Nones and nans as 'null'
+    expected_str = (
```
aditya-nambiar, line 86: this looks wierdly formatted? why is there a paranthesis ?

fennel/test_utils.py
```diff
@@ -32,3 +37,57 @@ def f(x: int, y: int) -> int:
...
+        "{"
+        + f'"num": [1.0, null], "word": ["abc", "def"], "bool": [null, false], "math": [3.14, 2.71828], "time": [{now_ms}, {now_ms}]'
+        + "}"
+    )
+    assert to_columnar_json(df3) == expected
+    assert to_columnar_json(df3, as_str=True) == expected_str
+    assert json.loads(to_columnar_json(df3, as_str=True)) == expected
```
aditya-nambiar, line 93: Can we add a complex type, like List of Dict or Dict of List and also a struct type to the test ?

aditya-nambiar (approved):

fennel/test_utils.py
zsid60: agree but this is what the linter likes. This is the result of running `black`

PR #272, opened 2023-08-21
Add spellcheck for docs

davidfaulkner12:
While I was blocked on onboarding I went through and did a quick spellcheck Github Action for our docs. There were quite a few random misspellings (as would inevitably happen without CI). Not a perfect solution, but at least our public docs won't have any more tipos like "Fennnel" :-)


davidfaulkner12:
An example of failing spellcheck: https://github.com/fennel-ai/client/actions/runs/5932533994/job/16086498008?pr=272

aditya-nambiar:
This is awesome!

aditya-nambiar (approved):

.wordlist.txt
```diff
@@ -0,0 +1,187 @@
...
+scalable
+schemas
+signup
+snowflakecomputing
+stateful
+str
+strftime
+struct
+tiering
+uid
```
aditya-nambiar, line 178: Can we keep it sorted across cases ?

.wordlist.txt
```diff
@@ -0,0 +1,187 @@
...
+protobuf
+protobufs
+quickstart
+realtime
+regex
+regexes
+repo
+runtime
+scalability
+scalable
```
aditya-nambiar, line 169: Can we add extractor, depends_on, version, inputs, ( all our sources maybe ? kinesis, bigquery etc )

davidfaulkner12: Most of those are already part of the dictionary -- I'm not quite sure why `scalable` and `scalability` aren't. It's just this text file (which doesn't even need to be sorted) so trivial to add things later.

.wordlist.txt
davidfaulkner12: Done.

PR #277, opened 2023-09-10
extract historical: Add functionality

aditya-nambiar:


fennel/client/client.py
```diff
@@ -400,6 +400,9 @@ def extract_historical_features(
         input_dataframe: Optional[pd.DataFrame] = None,
         input_bucket: Optional[str] = None,
         input_prefix: Optional[str] = None,
+        output_bucket: Optional[str] = None,
+        output_prefix: Optional[str] = None,
+        column_to_feature_map: Optional[Dict[str, Feature]] = None,
```
bothra90, line 405: It might be better to invert this to be a map from Feature -> str rather than the other way around. Rationale is people often use different feature ids for the same logical feature (rather than aliasing). This allows them to use the same column in the input as two features.

bothra90 (approved):

PR #279, opened 2023-09-15
Add select op

thaqibm:
Add tests for select


fennel/datasets/test_dataset.py
```diff
@@ -1489,7 +1491,97 @@ class B:
         @inputs(A)
         def from_a(cls, a: Dataset):
             x = a.rename({"a1": "b1"})
-            return x.drop(["a2", "a3"])
+            return x.select(["b1"])
```
nikhilgarg28, line 1494: Having to write a list in `select(['b1', 'b2'])` is a sub-optimal experience. Why not use *args so that user could write `select('b1', 'b2')`? 

We could also do this for drop too (maybe do *args but if there is only one element and that is list of strings, change args to be that list?)

fennel/datasets/datasets.py
```diff
@@ -229,6 +229,12 @@ def rename(self, columns: Dict[str, str]) -> _Node:
     def drop(self, columns: List[str]) -> _Node:
         return Drop(self, columns)
 
+    def select(self, columns: List[str]) -> _Node:
+        ts = self.dsschema().timestamp
+        # Keep the timestamp col
+        drop_cols = list(filter(lambda c : c not in columns and c != ts, self.dsschema().fields()))
+        return Drop(self, drop_cols)
```
nikhilgarg28, line 236: I thought we were going to pass protos back and use Drop in the backend? cc @bothra90

aditya-nambiar: We decided to make it a client side change. The code will still be visible to the user as part of the pipeline

fennel/datasets/test_dataset.py
thaqibm: Added the change for ``select``. 
drop uses the ``columns=[..]`` in lots of places.
Ex:
```py
        return dealer.transform(
            lambda df: df.drop(columns=["address", "cars"]),
            schema,
        )
```

aditya-nambiar: we could support both ?

aditya-nambiar:
Can you add some functionality tests in client_test/test_dataset.py and some invalid cases in dataset/test_invalid_dataset.py

aditya-nambiar (changes requested):
Also please add to docs. 
To preview the docs.
cd docs/
make up
localhost:3001 I think

fennel/datasets/test_dataset.py
```diff
@@ -1472,24 +1472,133 @@ def extract_info(df: pd.DataFrame) -> pd.DataFrame:
...
+            "operandId": "10340ca369826992acc29dc84b073c18",
+        },
+        "id": "0d52839b6fb94cde94dea24334ad9bce",
+        "isRoot": True,
+        "pipelineName": "from_a",
+    }
+
+    operator_req = sync_request.operators[2]
+    expected_operator_request = ParseDict(o, ds_proto.Operator())
+    print(operator_req)
```
aditya-nambiar, line 1597: can you remove the prints

fennel/datasets/datasets.py
```diff
@@ -226,8 +226,46 @@ def __add__(self, other):
...
+            cols = [*args]
+            if len(args) == 1 and isinstance(args[0], list):
+                cols = args[0]
+            return cols
+        else:
+            raise ValueError(
+                f"must specify either 'columns' or positional arguments to {name}."
+            )
+
+    def drop(self, *args, columns: Optional[List[str]] = None) -> _Node:
```
aditya-nambiar, line 252: Can you reorganize the code to put the public ones up and the __ ones below them

aditya-nambiar (approved):
Can you add it to changelog.md and increment the version in pyproject.toml before merging

PR #280, opened 2023-09-20
DEV-1815 batched derived lookups 

zsid60:
completes DEV-1815, DEV-1816, DEV-1818


fennel/featuresets/featureset.py
```diff
@@ -580,6 +576,57 @@ def _get_expectations(self):
             "Expectations are not yet supported for featuresets."
         )
 
+    def _generate_lookup_extractors(self):
+        """
+        Generates the derived lookup extractors for the features of this FS.
+        A single extractor is generated for all fields with the same (dataset, provider) pair.
+        """
```
nikhilgarg28, line 583: What happens if the user tries to change/remove a single (dataset, provider) pair later? Will we end up changing extractor for all others too?

I wonder if instead of literally deriving a unified extractor, we should do optimization at the backend during extraction? cc @zsid60 @aditya-nambiar @bothra90

aditya-nambiar: Its basically changing your extractor right? We would now generate 2 batch extractors rather than 1. 
 We should check if versioning permits it, and if it does then its the usual sync path right ? 

What is so special about this case ?

nikhilgarg28: What will be the version of this new modified extractor? 

Generally speaking, I think this method of merging extractors on client side is rather fragile in my view. This is so because the abstraction to auto derive feature extractors essentially gives one independent extractor to each feature and this approach is breaking that invariant.

zsid60: Version will have to be vectorized on the server side regardless of where the merge is happening.

As for where the merge should occur, I don't yet see why that invariant should be the invariant / how well is that invariant preserved if we merge on the server side. I also need to think a bit more about how to merge on the server side across extractors. Happy to chat more in person today along with @aditya-nambiar

PR #285, opened 2023-10-05
dropnull

thaqibm:


fennel/datasets/datasets.py
```diff
@@ -893,6 +903,17 @@ def wrap(c: Type[T]) -> Dataset:
...
 
 
+def fennel_is_optional(type_):
+    return (
+        typing.get_origin(type_) is Union
+        and type(None) is typing.get_args(type_)[1]
+    )
+
+
+def fennel_optional_inner(type_):
```
aditya-nambiar, line 913: I feel this function deserves a better name

fennel/datasets/datasets.py
```diff
@@ -1956,6 +1994,29 @@ def visitDrop(self, obj) -> DSSchema:
         output_schema.name = output_schema_name
         return output_schema
 
+    def visitDropNull(self, obj):
+        input_schema = copy.deepcopy(self.visit(obj.node))
+        output_schema_name = f"'[Pipeline:{self.pipeline_name}]->dropnull node'"
+        if obj.columns is None or len(obj.columns) == 0:
```
aditya-nambiar, line 2000: We should allow dropnull on the entire dataset too? Which would basically create dropnulls for any optional column in the dataset, if there are no optional columns then we can throw an error

aditya-nambiar (changes requested):

aditya-nambiar:
Can you add documentation and bump the version

aditya-nambiar (approved):

docs/pages/api-reference/operators.md
```diff
@@ -130,9 +130,24 @@ Fennel does not allow you to drop keys or timestamp columns from a dataset.
 
 <pre snippet="api-reference/operators_ref#drop"></pre>
 
+
+### Dropnull
+
+Fennel allows you to drop null values from columns of a dataset using the `dropnull` operator.
```
nikhilgarg28, line 136: We should mention that this also changes the type of the column from `Option[T]` to `T`.

docs/pages/api-reference/operators.md
```diff
@@ -130,9 +130,24 @@ Fennel does not allow you to drop keys or timestamp columns from a dataset.
...
+
+The `dropnull` operator has the following parameters:
+
+1. `columns: List[str]` - positional argument, that specifies the list of columns to apply dropnull on from the dataset.
```
nikhilgarg28, line 140: Can user leave arguments to `dropnull` empty? If so, worth documenting that here?

nikhilgarg28: We should also mention that this operator could take an actual list of names or simply names without list literal (not exactly like this, hopefully in better language). Also, please do the same for select and drop if we haven't already.

nikhilgarg28 (commented):
FYI - the recommended way to write commit messages is as follows:

<module>: <upto 72 chars in total of description in imperative form>

<more details>

Module can literally be the name of a module or logical module. In this case, 'operator' may be a good name. The summary should be imperative e.g. 'add dropnull, fix docs for select' vs 'added dropnull' or 'dropnull'.

PR #286, opened 2023-10-20
source: Support since param in sources

aditya-nambiar:


fennel/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
 
+## [0.18.10] - 2023-10-20
+- Add support for `starting_from` in S3 source.
```
nikhilgarg28, line 4: Suggest to call this `starting` instead of `starting_from`?

nikhilgarg28 (commented):
Requesting an experience improvement.

fennel/CHANGELOG.md
nikhilgarg28: Actually 'since' and 'until' may be even better? ('starting' is fine too but 'ending' doesn't roll off that well)

bothra90 (approved):

PR #289, opened 2023-11-06
Tests: Use utc timestamp everywhere

aditya-nambiar:


vinaykakade (approved):
Thanks, confirming that the tests work in my local workspace (which is set to IST).

PR #291, opened 2023-11-08
extract features: Support strings in extract features 

aditya-nambiar:


fennel/client/client.py
```diff
@@ -341,6 +341,13 @@ def extract_features(
         for input_feature in input_feature_list:
             if isinstance(input_feature, Feature):
                 input_feature_names.append(input_feature.fqn())
+            elif isinstance(input_feature, str):
+                if "." not in input_feature:
```
nikhilgarg28, line 345: I guess we should also verify that there are exactly two parts if split on "."?

fennel/client/client.py
```diff
@@ -356,13 +363,21 @@ def extract_features(
...
+        output_feature_name_to_type: Dict[str, Any] = {}
         for output_feature in output_feature_list:
             if isinstance(output_feature, Feature):
                 output_feature_names.append(output_feature.fqn())
                 output_feature_name_to_type[
                     output_feature.fqn()
                 ] = output_feature.dtype
+            elif isinstance(output_feature, str):
+                if "." not in output_feature:
+                    raise Exception(
```
nikhilgarg28, line 375: Same here?

fennel/test_lib/mock_client.py
```diff
@@ -426,6 +426,13 @@ def extract_features(
         for input_feature in input_feature_list:
             if isinstance(input_feature, Feature):
                 input_feature_names.append(input_feature.fqn_)
+            elif isinstance(input_feature, str):
+                if "." not in input_feature:
+                    raise Exception(
```
nikhilgarg28, line 431: Same here? Maybe have a function like `is_plausible_feature_fqn` and use that everywhere?

fennel/test_lib/mock_client.py
```diff
@@ -479,6 +486,13 @@ def extract_historical_features(
...
```
nikhilgarg28, line 491: here too

fennel/test_lib/mock_client.py
```diff
@@ -769,6 +783,13 @@ def _run_extractors(
                 output_df[output_feature.fqn_] = intermediate_data[
                     output_feature.fqn_
                 ]
+            elif isinstance(output_feature, str):
+                if "." not in output_feature:
+                    raise Exception(
```
nikhilgarg28, line 788: And here.

nikhilgarg28 (commented):
Leaving some comments

PR #296, opened 2023-11-09
client: raise an exception instead of printing an error

mohitreddy1996:


fennel/utils.py
```diff
@@ -76,7 +76,7 @@ def fennel_get_source(obj: Any) -> str:
 def check_response(response: requests.Response):  # type: ignore
     """Check the response from the server and raise an exception if the response is not OK"""
     if response.status_code != 200:
-        print(
+        raise Exception(
```
bothra90, line 79: @mohitreddy1996 : why do this? raising an exception causes the program to terminate with a stacktrace that users can't do much about. We recently made this change to go from throwing an exception -> print

mohitreddy1996: IMO we should definitely stop/crash the process if a call fails. The main motivation is that, say a sync call fails today, we print something on the terminal and later start executing next calls. e.g. few of the workflows we have seen in the past with Epifi or our load tests is we call sync, sleep for a bit, log data and then attempt to extract features. Say if the sync call fails now, rest of the calls will likely fail. Isn't it a better experience to stop the execution and prompt the user that something failed?

I am coming from CI/CD workflow using Github Actions where as part of a workflow, we sync datasets/features to a dev tier, if it passes, sync in staging and so on.. Sync call fails but I have no way to interrupt the failure. Raising an exception at least allows me to catch the exception and do something of my own (ignore it, handle it, reraise it etc).

aditya-nambiar:
cc @nikhilgarg28 if you have any thoughts

nikhilgarg28:
I think there are situations where raising an error makes sense and there are situations when it doesn't. My sense is that we should have a parameter to control this behavior (e.g. `throws` or hopefully something better). And the default should be for it to raise the exception. We could also tie this behavior with 'preview' kwarg - if preview is set, don't raise else do raise. 

Also, if/when we do raise, we should make sure that Exception contains the appropriate context needed to handle the sync failures.

fennel/utils.py
bothra90: I agree with what you're saying about interrupting the control flow when a sync error happens and require users to explicitly handle it. My concern has mostly been for the UX of seeing a stacktrace that doesn't make any sense to you. A better idea might be for us to catch these exception in the `sync` function and re-throw it so stack trace looks simpler. wdyt?

mohitreddy1996:
@aditya-nambiar what was the conclusion on this bit? I think both Abhay and Nikhil agree that throwing exception in these scenarios is good? But the what gets printed as part of exception was the open question?

aditya-nambiar:
Yeah we now throw exception always. What gets printed is the server error code and the server message

PR #297, opened 2023-11-10
extract: Handle codegen for extractors that dont depend on datasets 

aditya-nambiar:
This handles a corner case when an extractor does not depend on datasets, but the featureset does. The code gen for the extractor requires the featureset, which inturn requires the datasets.

Required only for compilation.

Also introduce a better error for looking up non keyed datasets

### DIFF
--- a/fennel/featuresets/featureset.py
+++ b/fennel/featuresets/featureset.py
@@ -480,6 +480,31 @@ class Featureset:
         self._expectation = self._get_expectations()
         propogate_fennel_attributes(featureset_cls, self)
 
+    def get_dataset_dependencies(self):
+        """
+        This function gets the list of datasets the Featureset depends upon.
+        This dependency is introduced by features that directly lookup a dataset
+        via the DS-FS route.
+
+        The motivation for this function is to help generated the required code, even
+        if an extractor does not depend on a dataset, but is part of a featureset which
+        has these kinds of dependencies.
+        """
+        depended_datasets = []
+        for f in self._features:
+            if (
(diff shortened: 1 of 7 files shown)


fennel/featuresets/featureset.py
```diff
@@ -480,6 +480,32 @@ def __init__(
...
+        This dependency is introduced by features that directly lookup a dataset
+        via the DS-FS route.
+
+        The motivation for this function is to help generated the required code, even
+        if an extractor does not depend on a dataset, but is part of a featureset which
+        has these kinds of dependencies.
+        """
+        depended_datasets = []
+        for f in self._features:
+            print("f", f)
```
nikhilgarg28, line 495: Stray?

nikhilgarg28 (approved):

fennel/test_lib/integration_client.py
```diff
@@ -11,8 +11,8 @@
     import pyarrow as pa
     from fennel_client_lib import RustClient  # type: ignore
     from fennel_dataset import lookup  # type: ignore
-except ImportError as e:
-    print(f"exception during import {e}")
+except ImportError:
+    pass
```
zsid60, line 15: Why not leave the print message?

aditya-nambiar: every time customers import fennel it is printing this exception, which looks wierd

zsid60: But are these import errors not fatal to using the int client?

PR #298, opened 2023-11-12
schema: Use pd types rather than python types

aditya-nambiar:
To keep things understandable. Primitive types, int, float, string and bool are pandas types. All complex types are still python types, cause they are also objects in the pandas world

1. Also cast data to the dataset schema while logging.
2. Also check if any input datasets to a pipeline were missed during a sync call


docs/examples/getting-started/quickstart.py
```diff
@@ -136,6 +136,7 @@ def myextractor(cls, ts: pd.Series, uids: pd.Series, sellers: pd.Series):
 ]
 df = pd.DataFrame(data, columns=columns)
 response = client.log("fennel_webhook", "Product", df)
+print(response.json())
```
nikhilgarg28, line 139: Stray?

nikhilgarg28 (commented):
Didn't follow much of the logic. But I'm happy to stamp if/when you want me to.

nikhilgarg28 (approved):

bothra90:
@aditya-nambiar : Just checking if we need to worry about backwards compatibility

aditya-nambiar:
@bothra90 the proto's are the same and the end user experience is the same, the format for keeping data in mock client has been changed

PR #300, opened 2023-11-12
auto extractor: Bug fixes for code gen and default values

aditya-nambiar:
Code gen - when you have an extractor that does requires a provider or an alias, the code for the those featuresets should be recursively available for any extractor defined in the given featureset.
Default - After assigning defaults, the type of series is still the type for None ( when the lookup was performed ) => object, set the right type so that schema validation downstream doesn't fail

### DIFF
--- a/fennel/client_tests/test_complex_autogen_extractor.py
+++ b/fennel/client_tests/test_complex_autogen_extractor.py
@@ -1,0 +1,341 @@
+import sys
+
+from datetime import datetime, timedelta
+from typing import Optional
+
+import pandas as pd
+import pytest
+
+import fennel
+
+from fennel import meta, Count, Window, featureset, feature, extractor
+from fennel.client import Client
+from fennel.lib.schema import inputs, oneof, outputs
+from fennel.lib.aggregate import Max, Min
+from fennel.sources import Webhook, S3, MySQL
+from fennel.datasets import dataset, field, pipeline, Dataset
+from fennel.sources import source
+from fennel.test_lib import MockClient, mock
+
+webhook = Webhook(name="fennel_webhook")
+
+__owner__ = "uber-data@eng.com"
+
+
+@dataset
+@source(webhook.endpoint("RiderDataset"), tier="local")
+class RiderDataset:
+    rider_id: int = field(key=True)
+    created: datetime = field(timestamp=True)
+    birthdate: datetime
+
+
+@dataset
+@source(webhook.endpoint("RiderCreditScoreDataset"), tier="local")
+class RiderCreditScoreDataset:
+    rider_id: int = field(key=True)
+    created: datetime
+    score: float
+
+
+@dataset
+@source(webhook.endpoint("CountryLicenseDataset"), tier="local")
+@meta(owner="data@eng.com")
+class CountryLicenseDataset:
+    rider_id: int = field(key=True)
+    created: datetime
+    country_code: str
+
+
+@dataset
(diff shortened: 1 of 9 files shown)


fennel/client_tests/test_complex_autogen_extractor.py
```diff
@@ -0,0 +1,231 @@
...
+        age_years = (datetime.now() - birthdate).dt.total_seconds() / (
+            60 * 60 * 24 * 365
+        )
+        age_years = age_years.astype(int)
+        return pd.DataFrame({"age_years": age_years})
+
+
+@mock
+def test_complex_auto_gen_extractors(client):
+    initial = client.sync(
```
nikhilgarg28, line 173: Maybe add a few asserts here where we try to sync with subset of these datasets/featuresets and all of them fail.

fennel/lib/to_proto/source_code.py
```diff
@@ -40,8 +50,36 @@ def get_featureset_core_code(featureset: Featureset) -> str:
     # If python version 3.8 or below add @feature decorator
     if sys.version_info < (3, 9):
         source_code = f"@featureset\n{dedent(source_code)}"
-    z = _remove_empty_lines(source_code)
-    return z
+    return _remove_empty_lines(source_code)
+
+
+def get_featureset_gen_code(
```
nikhilgarg28, line 56: Is this logic tested via some unit tests? I'm not seeing any tests where we don't sync a few indirect dependencies and get an error back?

aditya-nambiar: yeah tests were failing and i fixed those :p let me create invalid test cases

fennel/client_tests/test_complex_autogen_extractor.py
```diff
@@ -0,0 +1,293 @@
...
+        input_feature_list=[RequestFeatures0.rider_id],
+        output_feature_list=[RiderFeatures],
+        input_dataframe=pd.DataFrame({"RequestFeatures0.rider_id": [1]}),
+    )
+    assert extracted_df.shape[0] == 1
+    assert (
+        extracted_df["RiderFeatures.created"].iloc[0]
+        == rider_df["created"].iloc[0]
+    )
+    assert extracted_df["RiderFeatures.dl_state"].iloc[0] == "Unknown"
```
zsid60, line 293: would it be good to test some of the features here that have dependencies on the FS/DS features? Like is_us_dl and age_years?

fennel/client_tests/test_complex_autogen_extractor.py
```diff
@@ -0,0 +1,293 @@
...
+    ) -> pd.DataFrame:
+        age_years = (datetime.now() - birthdate).dt.total_seconds() / (
+            60 * 60 * 24 * 365
+        )
+        age_years = age_years.astype(int)
+        return pd.DataFrame({"age_years": age_years})
+
+
+@mock
+def test_complex_auto_gen_extractors(client):
```
zsid60, line 173: Can you add a comment here just describing what functionality we're trying to test here? There are a lot of Datasets / Featuresets in this test, and I get that right now it's probably very relevant to the Turo POC. But I think coming back to this in a few weeks, it'd be hard to know or remember what exactly is being tested here

nikhilgarg28 (approved):
I'm stamping for now though looks like @zsid60 has few more requests.

zsid60 (approved):

PR #301, opened 2023-11-13
join: Handle conflict in LHS value and RHS key

aditya-nambiar:


nikhilgarg28:
Just to confirm, if there is a column in left that has same name as non-key column of right, such that after join two columns will have the same name, we will still flag that as an error, right?

Also, not sure what the error message is right now but might be good to add a reference to using `rename` operator before joining.

aditya-nambiar:
Yes we do. We already have the preventions/errors in place. The issue this time was a tricky one

LHS key account id joined with the RHS key id, but LHS had id as a value field too.

So pandas made it id_x. 

The join is semantically correct since we should just drop the id from RHS after the join. So the PR only fixes the execution aspect

nikhilgarg28:
Understood, thanks for explaining.

bothra90 (approved):

PR #310, opened 2023-11-26
extractors: Add dtype validation for autogenerated extractors

aditya-nambiar:


nikhilgarg28 (approved):

fennel/featuresets/test_derived_extractors.py
```diff
@@ -57,7 +57,7 @@ class UserInfo:
             default="unspecified",
         )
         # lookup with meta
-        age_years: float = (
+        age_years: int = (
```
zsid60, line 60: In the dataset age is defined as a float so why this change?

fennel/featuresets/test_invalid_featureset.py
```diff
@@ -238,3 +245,114 @@ class UserInfo:
...
+    # age should be optional[int]
+    if sys.version_info < (3, 9):
+        assert (
+            str(e.value)
+            == "Feature `UserInfo.age` has type `<class 'int'>` but expectected type `typing.Union[int, NoneType]`"
+        )
+    else:
+        assert (
+            str(e.value)
+            == "Feature `UserInfo.age` has type `<class 'int'>` but expectected type `typing.Optional[int]`"
```
zsid60, line 283: I'm a bit confused on the point of this PR. In this example the dataset field is non optional and so intuitively to me the extracted feature would also be non optional. For values where the lookup doesn't return anything, this is what the default is for. So if you'll always have an int from the ds or a default int, what is this automatic relaxing to an optional type giving us / why are we doing it? Is it for the case when there is no default value (if so, why is this generalized to even when a default is present?)

fennel/featuresets/featureset.py
```diff
@@ -576,6 +582,60 @@ def _get_extractors(self) -> List[Extractor]:
         return extractors
 
     def _validate(self):
+        cls_module = inspect.getmodule(self.__fennel_original_cls__)
+        if cls_module is not None and hasattr(
+            cls_module, FENNEL_GEN_CODE_MARKER
```
zsid60, line 587: What does this do?

fennel/featuresets/test_derived_extractors.py
aditya-nambiar: No its a float

fennel/featuresets/test_invalid_featureset.py
aditya-nambiar: I dont get your question, this example does not have a default value, and hence age should be Optional[int], if you specify a default it should be in

fennel/featuresets/featureset.py
aditya-nambiar: This is an interesting issue I encountered. So basically we ship code. But when the shipped code runs, since it runs under exec mode, all types are strings. Hence for the shipping code typing validation breaks. However we dont need to run validation once it has already run by the user, so we disable the validation when executing shipped code

zsid60: Ahh I see

fennel/featuresets/test_derived_extractors.py
zsid60: Oh sorry I was looking at the User dataset not UserInfo

PR #311, opened 2023-11-26
Fix assign logging dtype checks

aditya-nambiar:
1. Cast issue - See comment in fennel/test_lib/test_utils.py
 2. log API in mock client takes df as param and prod client takes dataframe, making it df. This is a breaking change, but I dont think anyone uses the python log api in prod ?

### DIFF
--- a/fennel/client/client.py
+++ b/fennel/client/client.py
@@ -112,7 +112,7 @@ class Client:
         self,
         webhook: str,
         endpoint: str,
-        dataframe: pd.DataFrame,
+        df: pd.DataFrame,
         batch_size: int = 1000,
     ):
         """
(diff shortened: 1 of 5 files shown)


nikhilgarg28 (approved):

fennel/client/client.py
```diff
@@ -112,7 +112,7 @@ def log(
         self,
         webhook: str,
         endpoint: str,
-        dataframe: pd.DataFrame,
+        df: pd.DataFrame,
```
zsid60, line 115: Can you change the param name in the docs too?

zsid60 (approved):
LGTM w/ comment. Interesting catch.
Though i have no info on log being used or not in prod

PR #312, opened 2023-11-28
Add Creds to Extract Historical On the Client Side

zsid60:
State breaking change for extract historical

Still need to update documentation and test

### DIFF
--- a/fennel/CHANGELOG.md
+++ b/fennel/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## [0.19.0] - 2023-11-28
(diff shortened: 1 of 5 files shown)


aditya-nambiar:
One main reason for using the existing source in extract historical is we can now unify our creds through the system. So if the user defines the S3 source elsewhere ( in code now or console later ) and does S3.get(name="xyz").bucket(abc, prefix=1234) it will still work. Hence we need to reuse the code paths too, so send the source proto and use them in the backend to figure the creds

aditya-nambiar (approved):
Approving, since we plan to do the single source code path as a separate PR

PR #319, opened 2023-12-11
DEV-2363 change extract_historical API to send source protos 

zsid60:


zsid60:
Not ready to merge yet:
1) Need to implement the proto deser at the rest endpoint in Rust -- may require wrapping the whole extract historical request as a proto
2) needs to be e2e tested

PR #325, opened 2023-12-21
[DEV-2320] added capability to specify window in str for aggregate

nonibansal:
added capability to specify window in str for aggregate:

1. Added tests in datasets.
2. Added tests in client_tests.
3. Added document for aggregate dataset in docs for string window.

### DIFF
--- a/docs/examples/datasets/pipelines.py
+++ b/docs/examples/datasets/pipelines.py
@@ -364,15 +364,19 @@ class ActivityGroupedDataset:
     @pipeline(version=1)
     @inputs(Activity)
     def create_grouped_dataset(cls, activity: Dataset):
-        return activity.dropnull(
-            columns=["amount"]
-        ).groupby(
-            "user_id"
-        ).aggregate(
-            [
-                Sum(of="amount", window="forever", into_field="amount_forever"),
-                Sum(of="amount", window="1d", into_field="amount_1d"),
-            ]
+        return (
+            activity.dropnull(columns=["amount"])
+            .groupby("user_id")
+            .aggregate(
+                [
+                    Sum(
+                        of="amount",
+                        window="forever",
+                        into_field="amount_forever",
+                    ),
+                    Sum(of="amount", window="1d", into_field="amount_1d"),
+                ]
+            )
         )
 
 
--- a/docs/examples/datasets/pipelines.py
+++ b/docs/examples/datasets/pipelines.py
@@ -426,7 +426,6 @@ def test_str_window_aggregates(client):
             '{"transaction_amount": 99, "merchant_id": 1422}',
             one_day_ago,
         ],
-
     ]
     columns = ["user_id", "action_type", "amount", "metadata", "timestamp"]
     df = pd.DataFrame(data, columns=columns)
--- a/docs/examples/datasets/pipelines.py
+++ b/docs/examples/datasets/pipelines.py
@@ -350,3 +350,88 @@ def test_multiple_pipelines(client):
     )
     assert df.shape == (4, 4)
     assert found.sum() == 4
+
+
+# docsnip aggregate_with_str_window
+@meta(owner="data-eng-oncall@fennel.ai")
+@dataset
+class ActivityGroupedDataset:
+    timestamp: datetime
+    user_id: int = field(key=True)
+    amount_forever: float
+    amount_1d: float
+
+    @pipeline(version=1)
+    @inputs(Activity)
+    def create_grouped_dataset(cls, activity: Dataset):
+        return activity.dropnull(
+            columns=["amount"]
+        ).groupby(
+            "user_id"
+        ).aggregate(
+            [
+                Sum(of="amount", window="forever", into_field="amount_forever"),
+                Sum(of="amount", window="1d", into_field="amount_1d"),
+            ]
+        )
+
+
+# /docsnip
+
+
+@mock
+def test_str_window_aggregates(client):
+    # # Sync the dataset
+    client.sync(datasets=[Activity, ActivityGroupedDataset])
+    now = datetime.now()
+    minute_ago = now - timedelta(minutes=1)
+    one_day_ago = minute_ago - timedelta(days=1)
+    data = [
+        [
+            14325,
+            "report",
+            49,
+            '{"transaction_amount": 49, "merchant_id": 1322}',
+            minute_ago,
+        ],
+        [
+            14325,
+            "report",
+            None,
+            '{"transaction_amount": 49, "merchant_id": 1322}',
+            minute_ago,
+        ],
+        [
+            14325,
+            "atm_withdrawal",
+            99,
+            '{"location": "mumbai"}',
+            minute_ago,
+        ],
+        [
+            14325,
+            "report",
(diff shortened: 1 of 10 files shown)


docs/examples/datasets/pipelines.py
```diff
@@ -350,3 +350,91 @@ def test_multiple_pipelines(client):
     )
     assert df.shape == (4, 4)
     assert found.sum() == 4
+
+
+# docsnip aggregate_with_str_window
```
nikhilgarg28, line 355: Is some place in documentation using this new docsnip?

docs/examples/datasets/pipelines.py
```diff
@@ -350,3 +350,91 @@ def test_multiple_pipelines(client):
...
+@meta(owner="data-eng-oncall@fennel.ai")
```
nikhilgarg28, line 356: We recently added support for specifying `__owner__` at module level so that this doesn't need to be specified for each dataset. That makes the documentation feel slightly less verbose. 

Assuming this docsnip is going in docs, could you please do that here too?

docs/examples/datasets/pipelines.py
```diff
@@ -350,3 +350,91 @@ def test_multiple_pipelines(client):
...
+@dataset
+class ActivityGroupedDataset:
+    timestamp: datetime
+    user_id: int = field(key=True)
```
nikhilgarg28, line 360: Minor but could you please put key field at the top?

docs/examples/datasets/pipelines.py
```diff
@@ -350,3 +350,91 @@ def test_multiple_pipelines(client):
...
+    amount_forever: float
+    amount_1d: float
+
+    @pipeline(version=1)
+    @inputs(Activity)
+    def create_grouped_dataset(cls, activity: Dataset):
+        return (
+            activity.dropnull(columns=["amount"])
```
nikhilgarg28, line 368: IIRC, `dropnull` can be called without using kwargs e.g. `dropnull('amount')`. Could you please check if that's true? If so, could you please use that syntax here?

In general, we obsess about keeping as few keystrokes/words in the docs/examples as possible because that leads to a better developer experience eventually (after all people just copy/paste docs examples).

docs/examples/datasets/pipelines.py
```diff
@@ -350,3 +350,91 @@ def test_multiple_pipelines(client):
...
+            .aggregate(
+                [
+                    Sum(
+                        of="amount",
+                        window="forever",
+                        into_field="amount_forever",
+                    ),
+                    Sum(of="amount", window="1d", into_field="amount_1d"),
+                ]
+            )
```
nikhilgarg28, line 379: After a recent change, IIRC, aggregate can take *args without having to actually pass a list. That is, you can probably get rid of `[` and `]`. If you do that, hopefully indentation changes by few spaces and you can stuff in `Sum(...)` in the same line instead of splitting it over multiple lines. 

If it still doesn't fit in a single line, could you please use a slightly shorter field name e.g. `amount_forever` to `sum_forever` or simply `amount`. 

As you can see, we also obsess over text fitting in fewer lines in docs so that it feels simpler to read/write.

docs/examples/datasets/pipelines.py
```diff
@@ -350,3 +350,91 @@ def test_multiple_pipelines(client):
...
+        ],
+    ]
+    columns = ["user_id", "action_type", "amount", "metadata", "timestamp"]
+    df = pd.DataFrame(data, columns=columns)
+    response = client.log("fennel_webhook", "Activity", df)
+    assert response.status_code == requests.codes.OK, response.json()
+    # Only the mock client keyed datasets contains the lookup parameter to access the data.
+    user_ids = pd.Series([14325, 14326])
+    ts = pd.Series([now, now])
+    data, found = ActivityGroupedDataset.lookup(ts, user_id=user_ids)
```
nikhilgarg28, line 437: I think there is a `lookup` method on the client too?

nikhilgarg28: Btw we have a mechanism where we can run all client tests against a real server too. So if you use syntax/APIs specific to mock client, that makes it impossible to test that particular functionality end to end.

fennel/datasets/test_dataset.py
```diff
@@ -2580,3 +2580,67 @@ def pipeline2(cls, a: Dataset, b: Dataset):
...
+                        window="forever",
+                        into_field=str(cls.count),
+                    ),
+                    Average(
+                        of="age",
+                        window=Window("forever"),
+                        into_field=str(cls.avg_age),
+                    ),
+                ]
+            )
```
nikhilgarg28, line 2609: Try to fit aggregates in one line each by removing `[`

nikhilgarg28 (changes requested):
Requesting some changes. 

Also, could you please go back and change ALL the examples in documentation to now use this string based syntax instead of window based. That will hopefully cut down a few lines here and there but more importantly, that is how this change will be adopted by newer users - they will just look at the existing docs and copy it through.

docs/examples/datasets/pipelines.py
nonibansal: yup found many in pipelines so copies the syntax, should I remove it?

docs/examples/datasets/pipelines.py
nonibansal: Got it. Should I go ahead and remove all meta from existing datasets in the file?

docs/examples/datasets/pipelines.py
nonibansal: okie

docs/examples/datasets/pipelines.py
nonibansal: makes sense

docs/examples/datasets/pipelines.py
nonibansal: got it

fennel/datasets/test_dataset.py
nonibansal: got it

docs/examples/datasets/pipelines.py
nonibansal: Actually no, we can remove this and change all the existing ones to reflect the change, what say you?

nikhilgarg28: Yes, that sounds like the right thing to do here. Thanks!

docs/examples/datasets/pipelines.py
nikhilgarg28: You can as long as the documentation itself isn't selectively explaining the concept of ownership (in which case we want to keep it). Also, in some places, the description of code snippets refers to line numbers so if you remove this line, the line numbers may change - if that happens, you'd need to modify the documentation text as well.

nonibansal: got it

docs/examples/datasets/pipelines.py
nonibansal: got it, anyhow removed this from documentation.

nikhilgarg28 (approved):

PR #327, opened 2024-01-08
[DEV-2064] Added functionality to include fennel struct in extractor automatically

nonibansal:
This PR includes the functionality to automatically add a fennel struct in includes of the extractor responsible for the feature having fennel struct as dtype.


aditya-nambiar:
IMO If we are solving this problem we should solve it fully. The common case could well be that the datasets have structs and the features are primitive and after a lookup the user tries to fetch fields from the struct and another error will be thrown.

I think we should create a separate function to recursively find all struct dependencies and then add it to the generated code of the required module.

PR #330, opened 2024-01-09
[DEV-2096] Added checks to check if join fields are optional or not

nonibansal:
Added checks to check if join fields are optional or not in class SchemaValidator -> visitJoin.

BTW we didn't need to add this check because in schema validator, there's are already a check to compare schema on key on left and right datasets, as right datasets keys have to key field by definition, having optional on them in not allowed. Therefore, there's already an error throwing that schema not matched between left and right. 

Having this explicit, gives better error handling though.


fennel/datasets/datasets.py
```diff
@@ -1924,6 +1924,11 @@ def is_subset(subset: List[str], superset: List[str]) -> bool:
                 )
             # Check the schemas of the keys
             for key in obj.on:
+                if fennel_is_optional(left_schema.get_type(key)):
+                    raise TypeError(
+                        f"Optional typed fields are not allowed as Key field in left schema got `{key}` "
```
aditya-nambiar, line 1929: I dont think Key field is the right terminology since in a join the fields on the LHS can be non-key fields too. 

Something along the lines of - , Fields used in a join operation must not be optional, found x of type Optional[y] in Pipeline xyz ?

aditya-nambiar (changes requested):
Can you also increment the version in the toml file to create a new version of the client

fennel/datasets/datasets.py
nonibansal: makes sense, refracted the error statement

nonibansal: Also increased the version.

fennel/datasets/datasets.py
```diff
@@ -1924,6 +1921,12 @@ def is_subset(subset: List[str], superset: List[str]) -> bool:
                 )
             # Check the schemas of the keys
             for key in obj.on:
+                if fennel_is_optional(left_schema.get_type(key)):
+                    raise TypeError(
+                        f"Fields used in a join operation must not be optional in left schema, "
```
aditya-nambiar, line 1926: s/operation/operator

aditya-nambiar (approved):

fennel/datasets/datasets.py
nonibansal: done

PR #331, opened 2024-01-09
[DEV-2550] fix tests in social network

nonibansal:


fennel/client_tests/test_social_network.py
```diff
@@ -151,8 +150,6 @@ def extractor_category_view(
         )
 
 
-@pytest.mark.skip(reason="Failing in main currently: TODO: DEV-2550")
-@pytest.mark.slow
```
aditya-nambiar, line 155: Lets keep the @pytest.mark.slow marker, to make iteration speed faster while running tests locally

fennel/client_tests/test_social_network.py
```diff
@@ -62,7 +61,7 @@ class CityInfo:
     @inputs(UserInfo)
     def count_city_gender(cls, user_info: Dataset):
         return user_info.groupby(["city", "gender"]).aggregate(
-            [Count(window=Window("1y 5s"), into_field="count")]
+            [Count(window=Window("6y 8s"), into_field="count")]
```
aditya-nambiar, line 64: I believe we have removed the requirement to specify Window, so you can directly specify "6y 8s"

aditya-nambiar (approved):

fennel/client_tests/test_social_network.py
nonibansal: alright

fennel/client_tests/test_social_network.py
nonibansal: will change this.

PR #336, opened 2024-01-12
[DEV - 2699] Added Window operator and generated python stub for the proto

nonibansal:
Added Window operator and generated python stub for the proto


mohitreddy1996 (approved):
Did you bump up the client version for adding window operator? I guess this completes the functionality so the version should ideally happen now?

aditya-nambiar (approved):

PR #337, opened 2024-01-13
[DEV-2700] Added operator Window in datasets, serialize and schema validator and…

nonibansal:
Added operator Window in datasets, serialize and schema validator and respective tests.
Please first review https://github.com/fennel-ai/client/pull/336


fennel/datasets/datasets.py
```diff
@@ -723,6 +729,50 @@ def dsschema(self):
...
+        self.node.out_edges.append(self)
+
+    def signature(self):
+        return fhash(
+            self.node.signature(),
+            self.keys,
+            self.type,
+            self.gap,
+            self.field,
+        )
```
nikhilgarg28, line 760: Do we still use signature/fhash etc anywhere on the backend? If not, time to deprecate?

fennel/datasets/datasets.py
```diff
@@ -63,6 +63,7 @@
     get_python_type_from_pd,
     FENNEL_STRUCT_SRC_CODE,
     FENNEL_STRUCT_DEPENDENCIES_SRC_CODE,
+    WindowStruct,
```
nikhilgarg28, line 66: `WindowStruct` isn't nice at all. Let's just use `Window`?

fennel/datasets/datasets.py
```diff
@@ -723,6 +729,50 @@ def dsschema(self):
...
+        self,
+        node: _Node,
+        keys: List[str],
+        type: str,
+        gap: Duration,
+        field: str,
+    ):
+        super().__init__()
+        if len(keys) == 0:
+            raise ValueError("Must specify at least one key")
```
nikhilgarg28, line 743: Error should be better - at least mention this is for window operator. 
Let's also add a validation to ensure that only `tumble`, `sliding`, `session` are allowed. Any other window types are invalid. Further, for `tumble`, `sliding` we should raise a different error that these windows aren't implemented yet.

fennel/datasets/datasets.py
```diff
@@ -2144,3 +2199,23 @@ def visitFirst(self, obj) -> DSSchema:
...
         return output_schema
+
+    def visitWindow(self, obj) -> DSSchema:
+        output_schema = copy.deepcopy(obj.dsschema())
+        output_schema_name = f"'[Pipeline:{self.pipeline_name}]->window node'"
+
+        if len(output_schema.keys) == 0:
+            raise ValueError(
+                f"`group_by` before `window` in {self.pipeline_name} must specify at least one key"
+            )
```
nikhilgarg28, line 2210: Are we going to force that window must be preeceded by a groupby? I think that's unnecessary as long as it has some key columns. In that case, the error message should not refer to groupby. 

Also IIRC, groupby itself will fail if it had no keys.

fennel/datasets/test_schema_validator.py
```diff
@@ -1061,3 +1061,212 @@ def create_dataset(cls, activity: Dataset):
         str(e.value)
         == "Field `userid` is a key or timestamp field in schema of assign node input '[Dataset:RatingActivity]'. Value fields are: ['rating', 'movie']"
     )
+
```
nikhilgarg28, line 1064: Very good tests. Suggestion to add couple more (if not already added):

1. It's valid to use `window` without groupby provided its operand already has some keys
2. It's an error if the output dataset doesn't have field called `window` (if not overriden) and the overriden field otherwise
3. The type of the window field must be `WindowStruct`
4. It's valid to add operators after window (e.g. filter, transform)
5. It's valid to compute aggregate after window (e.g. avg window duration, which one of the customer is asking for)

nikhilgarg28 (commented):
Love the test coverage. Requesting a few more changes. I'm sure @aditya-nambiar is also reviewing this in parallel.

fennel/datasets/datasets.py
aditya-nambiar: +1 was telling @nonibansal the same. We do populate it and send it to the backend, but its pointless and we should get rid of it.

nonibansal: We can get rid on them in another PR, I will create a task for tracking this.
https://linear.app/fennel-ai/issue/DEV-2715/get-rid-of-signaturefhash-in-python-client

fennel/datasets/datasets.py
nonibansal: Yup this just a placeholder. I am thinking some names that we can use. Not in favour of Window though as then there will be three Window -> Window operator, Window ("5y") and then this Window Struct. Although we have deprecated the use of Window ("5y") and users won't be using Window operator proto nor class. So ideally we can do that.

fennel/datasets/datasets.py
nonibansal: Will refractor this.

fennel/datasets/datasets.py
nonibansal: I don't think groupby itself fails currently. Also the window operator I thought is similar to first, so made the code in such a way that window comes after groupby. If we want other way also I can do that, but if we do then I think we should also make first also similar. @aditya-nambiar your thoughts on this?

fennel/datasets/test_schema_validator.py
nonibansal: Will add tests 2, 3, 4, 5 about 1 I had a doubt I raised in previous comment also.

fennel/datasets/datasets.py
aditya-nambiar: We had an internal discussion and the plan is 
1. Rename the Window("5y") to Last("5y"), since we will anyways be kicking it out its fine.
2. .window can create a WindowOperator ( this is internal struct name so doesn't matter much )
3. The externally visible struct will be called Window.

Let me know what you think ?

fennel/datasets/datasets.py
aditya-nambiar: @nikhilgarg28 , currently we require aggregate and first to be preceded by groupby and we could follow similar suite for now, later we can refactor the code to make them independent, but lets proceed to keep the code change contained.

fennel/datasets/datasets.py
```diff
@@ -2144,3 +2199,23 @@ def visitFirst(self, obj) -> DSSchema:
...
         output_schema.name = f"'[Pipeline:{self.pipeline_name}]->first node'"
         return output_schema
+
+    def visitWindow(self, obj) -> DSSchema:
+        output_schema = copy.deepcopy(obj.dsschema())
+        output_schema_name = f"'[Pipeline:{self.pipeline_name}]->window node'"
+
+        if len(output_schema.keys) == 0:
+            raise ValueError(
+                f"`group_by` before `window` in {self.pipeline_name} must specify at least one key"
```
aditya-nambiar, line 2209: The current error syntax is to put any user specified variable in ` <> ` so {self.pipeline_name} should come in ``. We do this so that users can quickly read the error and find what pertains to them from a huge error message

fennel/datasets/datasets.py
```diff
@@ -2144,3 +2199,23 @@ def visitFirst(self, obj) -> DSSchema:
...
+            )
+        if obj.type not in ["session", "tumble", "sliding"]:
+            raise ValueError(
+                f"`type` in {output_schema_name} must be either `session` or "
```
aditya-nambiar, line 2213: user stuff like {output_schema_name} comes in ``

fennel/datasets/test_schema_validator.py
```diff
@@ -1061,3 +1061,204 @@ def create_dataset(cls, activity: Dataset):
...
+            user_id: str
+            window: WindowStruct
+            t: datetime
+
+            @pipeline(version=1)
+            @inputs(PageViewEvent)
+            def pipeline_window(cls, app_event: Dataset):
+                return app_event.groupby().window(type="session", gap="10m")
+
+    assert str(e.value) == """Must specify at least one key"""
```
aditya-nambiar, line 1088: This is error contains very little context, also considering moving invalid tests that are not schema related to test_invalid_datasets

fennel/datasets/test_schema_validator.py
aditya-nambiar: +1 A few more tests where Window is an intermediate dataset and there are further downstream datasets

aditya-nambiar: Also a check if any key is optional

nonibansal: For key ti be optional, I think schema validator for groupBy will handle that, I will write tests for the same.

fennel/datasets/datasets.py
aditya-nambiar: I would say to do in this one, I dont expect the surface area to be large since we have anyways deprecated Window("") syntax, and would keep the number of windows limited

fennel/datasets/datasets.py
nonibansal: Got it.

fennel/datasets/datasets.py
```diff
@@ -473,6 +474,11 @@ def first(self) -> _Node:
             self.keys = self.keys[0]  # type: ignore
         return First(self.node, list(self.keys))  # type: ignore
 
+    def window(self, type: str, gap: str, field: str = "window") -> _Node:
```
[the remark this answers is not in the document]
nonibansal: alright

fennel/datasets/datasets.py
```diff
@@ -723,6 +729,50 @@ def dsschema(self):
...
+        gap: Duration,
+        field: str,
+    ):
+        super().__init__()
+        if len(keys) == 0:
+            raise ValueError("Must specify at least one key")
+        keys.append(field)
+        self.keys = keys
+        self.type = type
+        self.gap = gap
```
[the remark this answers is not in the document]
nonibansal: nope, will remove this.

fennel/datasets/datasets.py
```diff
@@ -723,6 +729,50 @@ def dsschema(self):
         return input_schema
 
 
+class Window(_Node):
+    def __init__(
+        self,
+        node: _Node,
+        keys: List[str],
+        type: str,
```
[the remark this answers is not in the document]
nonibansal: alright.

fennel/datasets/datasets.py
```diff
@@ -723,6 +729,50 @@ def dsschema(self):
...
+    def dsschema(self):
+        input_schema = self.node.dsschema()
+        keys = {
+            f: input_schema.get_type(f) for f in self.keys if f != self.field
+        }
+        keys[self.field] = WindowStruct
+        values = {}
+        return DSSchema(
+            keys=keys,
+            values=values,  # type: ignore
```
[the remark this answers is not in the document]
nonibansal: done.

fennel/datasets/datasets.py
```diff
@@ -2144,3 +2199,23 @@ def visitFirst(self, obj) -> DSSchema:
...
+
+    def visitWindow(self, obj) -> DSSchema:
+        output_schema = copy.deepcopy(obj.dsschema())
+        output_schema_name = f"'[Pipeline:{self.pipeline_name}]->window node'"
+
+        if len(output_schema.keys) == 0:
+            raise ValueError(
+                f"`group_by` before `window` in {self.pipeline_name} must specify at least one key"
+            )
+        if obj.type not in ["session", "tumble", "sliding"]:
```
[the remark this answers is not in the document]
nonibansal: done

fennel/datasets/datasets.py
nonibansal: done,

fennel/datasets/datasets.py
nonibansal: done

fennel/datasets/test_schema_validator.py
nonibansal: added.

fennel/datasets/test_schema_validator.py
nonibansal: Can you share an example with of window operator with respect to invalid_dataset?

fennel/lib/last/last.py
```diff
@@ -19,7 +19,7 @@
 # ------------------------------------------------------------------------------
 
 
-class Window(BaseModel):
+class Last(BaseModel):
```
nikhilgarg28, line 22: Having `Last` is great, but I don't think we can remove old `Window` just like that - else it's a breaking change for existing users?

I'd suggest keeping old `Window` around for now (it will have different path from new `Window` so won't collide) + all documentation can change to using either `Last` or no constructor. 

The next time we do a breaking change (say in `1.0.0` launch), we can then remove the old `Window`. cc @aditya-nambiar for thoughts.

nonibansal: Removing window will be a breaking change.

I was thinking along the same lines, anyhow this change has to be a major version too, for now I will add the window, but I will change the docs to remove any mention wither Window or Last.

nikhilgarg28: Sounds good. Yeah, we are planning a `1.0` alpha release at the end of Q1A - we could make a batch of breaking changes to the API in that one.

fennel/datasets/datasets.py
```diff
@@ -473,6 +475,13 @@ def first(self) -> _Node:
             self.keys = self.keys[0]  # type: ignore
         return First(self.node, list(self.keys))  # type: ignore
 
+    def window(self, type: str, gap: str, field: str) -> _Node:
```
aditya-nambiar, line 478: the type is not str but the enum right ?

fennel/datasets/datasets.py
```diff
@@ -2144,3 +2216,18 @@ def visitFirst(self, obj) -> DSSchema:
             )
         output_schema.name = f"'[Pipeline:{self.pipeline_name}]->first node'"
         return output_schema
+
+    def visitWindow(self, obj) -> DSSchema:
+        output_schema = copy.deepcopy(obj.dsschema())
+        output_schema_name = f"'[Pipeline:{self.pipeline_name}]->window node'"
+
```
aditya-nambiar, line 2223: Lets add a schema check for gap? I think we rely on the to proto to currently ?

fennel/datasets/datasets.py
```diff
@@ -473,6 +475,13 @@ def first(self) -> _Node:
             self.keys = self.keys[0]  # type: ignore
         return First(self.node, list(self.keys))  # type: ignore
 
+    def window(self, type: str, gap: str, field: str) -> _Node:
+        if len(self.keys) == 1 and isinstance(self.keys[0], list):
+            self.keys = self.keys[0]  # type: ignore
+        return WindowOperator(
```
aditya-nambiar, line 481: Can you also add a check that field should not be among the list of columns already present ?

aditya-nambiar (approved):
Bump up the client version ?

fennel/datasets/datasets.py
nikhilgarg28: I actually thought `str` might be better because enums don't feel pythonic. For instance, `how` parameter of join takes a string, not enum (and same in pandas too). 

In fact, I've asked @zsid60 to change one of the enums in Kinesis connector to use str instead of enums.

nonibansal: I think initiation should be from str, we are internally converting str to Enum, but if method has Enum, then that's an extra import user has to do.

fennel/datasets/datasets.py
nonibansal: That is the behavior in join operator but for Window I have called the function which converts string to Timedelta in initiation itself, it will throw the at that time only. Or I can call the function in Schema Validator but we shouldn't assign obj params in Schema Validator, so didn't add there. Anyhow I have written the test for this exactly.

fennel/datasets/datasets.py
nonibansal: Yup, will add this.

PR #340, opened 2024-01-18
DEV-2612 kinesis client changes

zsid60:
image

### DIFF
--- a/fennel/lib/to_proto/to_proto.py
+++ b/fennel/lib/to_proto/to_proto.py
@@ -41,6 +41,7 @@ from fennel.lib.to_proto.source_code import (
     get_featureset_gen_code,
 )
 from fennel.lib.to_proto.source_code import to_includes_proto
+from fennel.sources import kinesis
 from fennel.utils import fennel_get_source
 
 
@@ -1131,7 +1132,6 @@ def _kinesis_conn_to_source_proto(
         db=ext_db,
         stream_arn=connector.stream_arn,
         init_position=connector.init_position,
-        init_timestamp=connector.init_timestamp,
         format=connector.format,
     )
     if connector.starting_from is not None:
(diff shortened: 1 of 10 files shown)


zsid60:
@nikhilgarg28 I verified that the datetime -> proto.Timestamp conversion behaves as expected with timezones

nikhilgarg28 (approved):

fennel/lib/to_proto/to_proto.py
```diff
@@ -1173,25 +1173,21 @@ def _kinesis_to_ext_db_proto(
 def _kinesis_to_ext_table_proto(
     db: connector_proto.ExtDatabase,
     stream_arn: str,
-    init_position: sources.InitPosition,
-    init_timestamp: Optional[datetime] = None,
+    init_position: str | kinesis.at_timestamp,
     format: str = "json",
```
mohitreddy1996, line 1177: should we remove default value from here too and add a validation that format must be json?

mohitreddy1996 (approved):

fennel/lib/to_proto/to_proto.py
mohitreddy1996: We only need to make the argument on the public API optional and have a default value. As the value propagates down the code to protos, we shouldn't have default values? But rather add validation that it is not json?

zsid60: the validation is already present and tested but good catch, will remove this here

mohitreddy1996: I don't think there is a validation added here in this method (to_proto.py file for this). Please add one and add an unit test for it in test_to_proto.py (this future proofs that if there is a different caller for this, which there could easily be, fails if they pass an invalid format).

PR #342, opened 2024-01-18
[DEV-2653] Fixed DOB calculation logic in documentation and tests

nonibansal:
Using ts instead of datetime.now() to make extract_historical use as of time instead of always using current time.


aditya-nambiar (approved):
Maybe add a comment too on why we should use ts and not datetime.now() to clarify to users

PR #343, opened 2024-01-18
[DEV-2698] Added documentation of Window operator in the API Reference

nonibansal:


docs/pages/api-reference/operators.md
```diff
@@ -236,4 +236,25 @@ The `assign` operator has the following parameters:
 Fennel does not allow you to assign keys or timestamp columns from a dataset.
 :::
 
+### Groupby/ Window
 
+Fennel allows you to groupby and get a stream of windowed events for each group using the `window` operator.
+A great example of windowed events are sessions, where in we create window of events occurring together.
```
aditya-nambiar, line 242: In Fennel, the `window` operator enables grouping events into windowed streams based on specified criteria. A common application of this operator is in creating sessions, where it forms a window comprising events that occur in a session.
The `window` operator must be preceded by `groupby` to identify the fields that will be used for grouping the events before they are windowed.

docs/pages/api-reference/operators.md
```diff
@@ -236,4 +236,25 @@ The `assign` operator has the following parameters:
...
+
+1. ``type: str``: positional argument, that specifies the type of windows to create. Allowed types
+   are `session`, `tumble` and `sliding`.
+2. ``gap: str``: positional argument, that specifies the maximum time gap between two consecutive events for clubbing them into a single window.
+3. ``field: str``: positional argument, that specifies the name of the keyed window field. The window field will contain following information: begin time, end time and number of events in a particular window.
+
+<pre snippet="api-reference/operators_ref#window"></pre>
+
+:::info
+The `tumble` and `sliding` window type in the `window` operator are still in development phase.
```
aditya-nambiar, line 259: can remove "in the `window` operator" since it is inside this section

docs/pages/api-reference/operators.md
nonibansal: made the change.

docs/pages/api-reference/operators.md
nonibansal: made the change

docs/pages/api-reference/operators.md
```diff
@@ -236,4 +236,24 @@ The `assign` operator has the following parameters:
...
+The `window` operator must be preceded by `groupby` to identify the fields that will be used for grouping the events before they are windowed.
+
+`groupby` takes the following parameters:
+
+1. `*args: str` - positional arguments that specify the list of fields to group by.
+
+The `window` operator takes the following parameters:
+
+1. ``type: str``: positional argument, that specifies the type of windows to create. Allowed types
+   are `session`, `tumble` and `sliding`.
```
nikhilgarg28, line 251: Should we mention that only `session` windows are supported as of right now?

docs/pages/api-reference/operators.md
```diff
@@ -236,4 +236,24 @@ The `assign` operator has the following parameters:
...
+2. ``gap: str``: positional argument, that specifies the maximum time gap between two consecutive events for clubbing them into a single window.
```
nikhilgarg28, line 252: This should not be a positional argument but only a kwarg - reason being that for 'tumble' windows, there will be no gap.

aditya-nambiar:
Also @nonibansal  can you add the Window struct scheme to the documentation so that users know what they contain

docs/pages/api-reference/operators.md
```diff
@@ -236,4 +236,24 @@ The `assign` operator has the following parameters:
...
+3. ``field: str``: positional argument, that specifies the name of the keyed window field. The window field will contain following information: begin time, end time and number of events in a particular window.
```
nikhilgarg28, line 253: Should be a kwarg, not positional arg. Also please specify that a field with name will be created in the output dataset with type `Window` with a link to some info about the `Window` struct (or put that in-line itself).

PR #346, opened 2024-01-18
Mock client raises an exception rather than returning 400 response in case of failure

aditya-nambiar:


docs/examples/api-reference/data-types.py
```diff
@@ -84,5 +84,5 @@ def test_restrictions(client):
             "timestamp": now,
         },
     ]
-    res = client.log("fennel_webhook", "UserInfoDataset", pd.DataFrame(data))
-    assert res.status_code == 400, res.json()
```
nikhilgarg28, line 88: Should we still add assertion about the code of the exception being 400?

aditya-nambiar: there is no code, pure python exception

nikhilgarg28 (changes requested):
Looks great. Just requesting one small change - somehow don't lose the error code from exceptions. Maybe add `FennelException` class which holds both `code` and `message` or something?

PR #352, opened 2024-01-22
[DEV-2728] Added capability to do as-of in method lookup in the client

nonibansal:
Added capability to do as-of in method lookup in the client

### DIFF
--- a/docs/pages/api-reference/client.md
+++ b/docs/pages/api-reference/client.md
@@ -186,7 +186,7 @@ The `lookup` method allows users to lookup values from keyed datasets.&#x20;
 * `dataset_name: str`: The name of the dataset against which we want to lookup values.
 * `keys: List[Dict[str, Any]]` : List of dict where each dict can be thought of as one entity for which we want to lookup. 
 * `fields: List[str]`: List of field names in the dataset that we want to lookup.
-* `timestamps: List[Union[int, str, datetime]]` - Pass this argument if we want to lookup as of a specific time. The length of this list should be exactly equal to length of keys argument. Here string can be representation of datetime supported by pd.to_datetime() and int can be unix time in either seconds, milliseconds or microseconds.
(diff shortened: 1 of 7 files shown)


fennel/client/client.py
```diff
@@ -705,10 +705,25 @@ def extract_historical_cancel_request(self, request_id):
         )
 
     def lookup(
-        self, dataset_name: str, keys: List[Dict[str, Any]], fields: List[str]
+        self,
+        dataset_name: str,
+        keys: List[Dict[str, Any]],
+        fields: List[str],
+        timestamps: List[Union[int, str]] = None,
```
aditya-nambiar, line 712: Why is datetime type not there ?

aditya-nambiar (changes requested):
Could you add this to the docs too

fennel/client/client.py
nonibansal: added that, forgot about that.

docs/pages/api-reference/client.md
```diff
@@ -175,3 +175,35 @@ The response format of this function and the `extract_historical` function are i
...
+### **lookup**
+
+The `lookup` method allows users to lookup values from keyed datasets.&#x20;
+
+**Arguments:**
+
+* `dataset_name: str`: The name of the dataset against which we want to lookup values.
+* `keys: List[Dict[str, Any]]` : List of dict where each dict can be thought of as one entity for which we want to lookup. 
+* `fields: List[str]`: List of field names in the dataset that we want to lookup.
+* `timestamps: List[Union[int, str, datetime]]` - Pass this argument if we want to lookup as of a specific time. The length of this list should be exactly equal to length of keys argument.
```
aditya-nambiar, line 189: Can we add what the str and ints can be ?

aditya-nambiar (approved):

docs/pages/api-reference/client.md
nonibansal: added

PR #356, opened 2024-01-23
API's for branch creation/deletion/list

aditya-nambiar:


fennel/client/client.py
```diff
@@ -409,42 +243,62 @@ def extract(
...
+        Parameters:
+        name (str): The name of the branch to delete.
+
+        Returns:
+        Dict[str, Any]: A dictionary containing the branch name and the status of the request.
+        """
+        req = {"branch_name": name}
+        return self._post_json("{}/branch/delete".format(V1_API), req)
+
+    def list_branches(self, user: Optional[str] = None):
```
nikhilgarg28, line 288: This operation shouldn't take `user` - instead it should return branches based on the bearer of the token.

fennel/client/client.py
```diff
@@ -409,42 +243,62 @@ def extract(
...
+    def clone_branch(self, name: str, from_branch: str):
+        """
+        Clone a branch from another branch.
+
+        Parameters:
+        name (str): The name of the branch to create.
+        from_branch (str): The name of the branch to clone from.
+
+        Returns:
+        Dict[str, Any]: A dictionary containing the branch name and the status of the request.
```
nikhilgarg28, line 270: Why it is returning a dictionary? Do you imaging extending this method to clone many branches at once someday? If not, I'd imagine it should return just the `status` of that one branch that was requested?

fennel/client/client.py
```diff
@@ -409,42 +243,62 @@ def extract(
...
-        input_feature_list: List[Union[Feature, str]],
-        output_feature_list: List[Union[Feature, Featureset, str]],
-        input_dataframe: pd.DataFrame,
-        log: bool = False,
-        workflow: Optional[str] = None,
-        sampling_rate: Optional[float] = None,
-    ) -> Union[pd.DataFrame, pd.Series]:
+    # ----------------------- Branch API's -----------------------------------
+
+    def empty_branch(self, name: str):
```
nikhilgarg28, line 248: All other API methods start with a verb except this one. Consider using `create_branch` or `init_branch`

fennel/client/client.py
```diff
@@ -409,42 +243,62 @@ def extract(
...
-        output_feature_list (List[Union[Feature, Featureset, str]]): List of feature or featureset objects or fully qualified feature names (when providing a str) to compute.
-        input_dataframe (pd.DataFrame): Dataframe containing the input features.
-        log (bool): Boolean which indicates if the extracted features should also be logged (for log-and-wait approach to training data generation). Default is False.
-        workflow (Optional[str]): The name of the workflow associated with the feature extraction. Only relevant when log is set to True.
-        sampling_rate (float): The rate at which feature data should be sampled before logging. Only relevant when log is set to True. The default value is 1.0.
+        name (str): The name of the branch to create.
 
         Returns:
-        Union[pd.DataFrame, pd.Series]: Pandas dataframe or series containing the output features.
+        Dict[str, Any]: A dictionary containing the branch name and the status of the request.
```
nikhilgarg28, line 256: Why is it returning a dictionary instead of just a single status?

nikhilgarg28: Worth documenting what is the structure of this `Status` object? Can I just get code out of it? Something else too?

fennel/client/client.py
```diff
@@ -409,42 +243,62 @@ def extract(
...
+
+    def list_branches(self, user: Optional[str] = None):
+        """
+        List all branches that the user has access to.
+
+        Parameters:
+        user (Optional[str]): The name of the user to list branches for. If not provided, the branches for all users will be listed.
+
+        Returns:
+        Dict[str, Any]: A dictionary containing the branch name and the status of the request.
```
nikhilgarg28, line 296: I don't know what this means. I suppose this method should return a list of branch metadata objects?

nikhilgarg28 (changes requested):
Requesting some changes.

PR #357, opened 2024-01-23
Added schema checks for values in list, dict, struct

nonibansal:
Added schema checks for values in list, dict, struct


aditya-nambiar:
We need to add a lot more tests to this. List[Struct], Dict[str, List[Float]], Optional[X], Struct containing a list field, Dict[str, Struct] etc

PR #363, opened 2024-01-30
[DEV-2634] Update client to be compatible with window backend change.

petrpan26:
Fix test and regenerate proto for client


fennel/datasets/datasets.py
```diff
@@ -764,19 +777,41 @@ def __init__(
...
+        if duration is None:
+            if type == WindowType.Hopping or type == WindowType.Tumbling:
+                raise ValueError(
+                    "'hopping window' or 'tumbling window' must specify duration"
+                )
+
+        if gap is None:
+            if type == WindowType.Sessionize:
+                raise ValueError("'sessionize window' must specify gap")
+
```
nikhilgarg28, line 803: Nit but I think it will be cleaner if you did validation by first matching on window type. That will allow you to validate that all required parameters are provided AND none of the parameters that don't make sense aren't provided.

nikhilgarg28: Could you please add some tests for these scenarios too?

nikhilgarg28 (changes requested):

fennel/datasets/datasets.py
```diff
@@ -764,19 +777,43 @@ def __init__(
...
+        gap: Optional[Duration],
+        duration: Optional[Duration],
+        stride: Optional[Duration],
     ):
         super().__init__()
         if len(keys) == 0:
             raise ValueError(
                 "'group_by' before 'window' must specify at least one key"
             )
+        if type == WindowType.Hopping:
```
nikhilgarg28, line 790: Could you please add negative checks too i.e. if type is Hopping, gap isn't provided. Or if it's tumbling or session, stride isn't provided.

fennel/datasets/datasets.py
```diff
@@ -2248,7 +2285,7 @@ def visitWindow(self, obj) -> DSSchema:
             raise ValueError(
                 f"'group_by' before 'window' in `{self.pipeline_name}` must specify at least one key"
             )
-        if obj.type in [WindowType.Tumbling, WindowType.Sliding]:
+        if obj.type in [WindowType.Hopping, WindowType.Tumbling]:
             raise NotImplementedError(
                 f"`{obj.type.value}` type not yet implemented in the window operator"
             )
```
nikhilgarg28, line 2291: Maybe remove this now since it will soon be ready?

fennel/lib/schema/schema.py
```diff
@@ -386,6 +386,8 @@ def to_proto(self):
         )
 
 
+# Window represent a time window that encapsulates event in [begin, end)
+# (e.g x in w if w.begin <= x < w.end)
```
nikhilgarg28, line 390: Could you please move this to docstring of class vs a comment on top?

fennel/lib/schema/schema.py
```diff
@@ -388,6 +388,9 @@ def to_proto(self):
 
 @struct
 class Window:
+    """Window represent a time window that encapsulates event in [begin, end)
```
nikhilgarg28, line 391: Not requesting change for now but JFYI, this docstring doesn't align with official style guide: https://peps.python.org/pep-0257/ (Also read https://peps.python.org/pep-0008/ someday).

fennel/datasets/test_schema_validator.py
```diff
@@ -1367,24 +1369,176 @@ class PageViewEvent:
...
 
     assert (
         str(e.value)
-        == """Window field name `page_id` in `'[Pipeline:pipeline_window]->window node'` must be different from non keyed fields in `'[Dataset:PageViewEvent]'`"""
+        == """'sessionize window' doesn't allow duration parameter"""
     )
+
+    with pytest.raises(ValueError) as e:
+
+        @meta(owner="nitin@fennel.ai")
```
nikhilgarg28, line 1523: JFYI - you could write `__owner__` at the top of module after which this doesn't need to be provided for each dataset/featureset.

nikhilgarg28 (approved):

PR #367, opened 2024-01-30
DEV-3032, DEV-3033 Introduce S3 `path` parameter 

zsid60:
Path is split into prefix (a fixed static string, can be multiple levels) and a suffix (supports the wildcards `*`, `**`, `*.{ext}`, or `strptime` datetime specifiers.

The splitting and validations are done on the client side. So `path` is fully a client construct, while the server gets a prefix and suffix

Docs rendering:
image


fennel/sources/test_invalid_sources.py
```diff
@@ -327,6 +327,36 @@ def test_invalid_s3_format():
...
+        s3.bucket(
+            bucket_name="bucket", prefix="prefix", path="prefix/suffix/*.json"
+        )
+    assert "path and prefix cannot be specified together" == str(e.value)
+
+    with pytest.raises(AttributeError) as e:
+        s3.bucket(bucket_name="bucket")
+    assert "either path or prefix must be specified" == str(e.value)
+
+    invalid_path_cases = [
```
nikhilgarg28, line 342: Good test cases.

nikhilgarg28 (approved):

PR #369, opened 2024-01-31
s3 path: allow equals in the source path

mohitreddy1996:


zsid60:
👍

zsid60:
We could just use the list here excluding *

https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html#object-key-guidelines

PR #370, opened 2024-02-01
[DEV-3075] Add mock client logic for tumbling and hopping and unit test

petrpan26:
Add tumbling and hopping window capability for mock client as well as adding unit tests to test it.


aditya-nambiar:
Code looks good to me. However there are 2 kinds of tests missing.
1. Proto generation test i.e are we creating the correct protos, check test_dataset.py for a bunch of proto validation.
2. There are very less invalid tests for eg the error thrown if i give a wrong window type, gap is negative, stride > duration etc
3. Can we also add a test where we have a hopping window with stride = duration

petrpan26:
2) I don't think we can even define negative time because it's not valid, Same as other window in aggregate
3) Why is this test needed ?

PR #371, opened 2024-02-02
Introduce bounded and idleness on data sources

saiharshavellanki:
- Introduce bounded and idleness fields on data sources to handle sources with limited data that are primarily used for backfill
- Added unit tests to verify things work as expected
- Updated the documentation

Screenshot 2024-03-05 at 5 09 36 PM

Screenshot 2024-03-05 at 5 09 42 PM

Screenshot 2024-03-05 at 5 10 12 PM

### DIFF
--- a/docs/pages/concepts/source.md
+++ b/docs/pages/concepts/source.md
@@ -28,6 +28,12 @@ And that's it - once this is written, `UserLocation` dataset will start
 mirroring your postgres table `user`and will update as the underlying Postgres 
 table updates.
 
+### Bounded
+The `bounded` attribute in the source indicates whether the source possesses a finite amount of data that does not expand. It is assigned a value of True when the source has limited data, with the default value being False.
+
+### Idleness
+The idleness parameter, when non-null, signifies that a bounded source is expected to be closed after a specified duration. For unbounded sources, it is advised not to set this parameter.
+
 ## Schema Matching
 
 Fennel has a strong typing system and all ingested data is evaluated against
--- a/docs/pages/concepts/source.md
+++ b/docs/pages/concepts/source.md
@@ -28,12 +28,6 @@ And that's it - once this is written, `UserLocation` dataset will start
 mirroring your postgres table `user`and will update as the underlying Postgres 
 table updates.
 
-### Bounded
-The `bounded` attribute in the source indicates whether the source possesses a finite amount of data that does not expand. It is assigned a value of True when the source has limited data, with the default value being False.
-
-### Idleness
-The idleness parameter, when non-null, signifies that a bounded source is expected to be closed after a specified duration. For unbounded sources, it is advised not to set this parameter.
-
 ## Schema Matching
 
 Fennel has a strong typing system and all ingested data is evaluated against
@@ -152,6 +146,21 @@ When deriving datasets that are unions of batch and realtime sources, `until` an
 to switch between sources. For example, read from s3 `until` a timestamp, and from that timestamp onwards, read from kafka
 using `since`.  
 
+### Bounded
+The `bounded` field in the source indicates whether the source possesses  a finite amount of data that does not expand.
+It is assigned a value of True when the source has limited data, with the default value being False.
+
+This ensures correct watermark values are propagated to the downstream pipelines. Without this, say an unbounded source
+with no data flow will cause the watermarks to be stuck and will have cost and performance implications.
+
+### Idleness
+The `idleness` field, when non-null, signifies that a bounded source is 
+expected to be closed after a specified duration. It is invalid if set for
+unbounded sources
+
+Let's see an example on how to use bounded and idleness parameters:
+
+<pre snippet="concepts/introduction#bounded_idleness"></pre>
 
 ## Load Impact of Sources
 
--- a/fennel/client_tests/test_bounded_idleness.py
+++ b/fennel/client_tests/test_bounded_idleness.py
@@ -18,7 +18,7 @@ __owner__ = "saiharsha@fennel.ai"
     cdc="append",
     disorder="1d",
     bounded=True,
-    idleness="4s",
+    idleness="10s",
 )
 @dataset
 class BoundedClicksDS:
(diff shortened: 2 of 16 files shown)


docs/pages/concepts/source.md
```diff
@@ -26,6 +26,11 @@ table named `user` within the Postgres database. And that's it - once this
 is written, `UserLocation` dataset will start mirroring your postgres table
 `user`and will update as the underlying Postgres table updates.
 
+### Bounded
+The `bounded` attribute in the source indicates whether the source possesses a finite amount of data that does not expand. It is assigned a value of True when the source has limited data, with the default value being False.
+
+### Idleness
+The idleness parameter, when non-null, signifies that a bounded source is expected to be closed after a specified duration. For unbounded sources, it is advised not to set this parameter.
```
mohitreddy1996, line 33: "For unbounded sources, it is advised not to set this parameter."

We would fail for unbounded sources if idleness is set right? So can you reword it to something like "invalid if set for unbounded sources"

fennel/sources/test_sources.py
```diff
@@ -1247,3 +1247,148 @@ class UserInfoDataset:
         assert (
             suffix == expected_suffix
         ), f"Expected suffix: {expected_suffix}, got: {suffix}"
```
mohitreddy1996, line 1249: can you assert on an unbounded source as well that bounded is set to False and idleness is set to None?

docs/pages/concepts/source.md
```diff
@@ -26,6 +26,11 @@ table named `user` within the Postgres database. And that's it - once this
 is written, `UserLocation` dataset will start mirroring your postgres table
 `user`and will update as the underlying Postgres table updates.
 
+### Bounded
```
mohitreddy1996, line 29: I think we should also add an example of how bounded and idleness will most likely be used i.e. 

One batch source with bounded and idleness set. Also a comment saying something like 'after initial ingestion, this data source is considered/marked as closed due to no data flow for `idleness` time period'.
One stream source 

There is a small pipeline which say does filter first on each individually say on timestamp column so avoid duplicate data. Something like -

@pipeline(ds_batch, ds_stream)
def pipe():
   ds_batch = ds_batch.filter('ts' < '2024-02-20')
   ds_stream = ds_stream.filter('ts' >= '2024-02-20')
    return ds_batch + ds_stream

docs/pages/concepts/source.md
```diff
@@ -26,6 +26,11 @@ table named `user` within the Postgres database. And that's it - once this
...
+The `bounded` attribute in the source indicates whether the source possesses a finite amount of data that does not expand. It is assigned a value of True when the source has limited data, with the default value being False.
```
mohitreddy1996, line 30: I think we should also share the idea behind this. Why does the user care if the source is bounded or not? 

The reason is for watermarks. I think we should write about it, but not too much in detail. We should say something in the lines of -
this ensures correct watermark values are propagated to the downstream pipelines. Without this, say an unbounded source with no data flow will cause the watermarks to be stuck and will have cost and performance implications.

mohitreddy1996 (commented):
We have a mock_client which is like an in-memory processing engine. We try to simulate whatever we do in the backend in this as well. This unlocks a very powerful use case of the users - unit testing their fennel based code

- https://github.com/fennel-ai/client/blob/9032411585d46f382bd558081790626e2d2eff2e/fennel/test_lib/mock_client.py#L4

We will need to add support for closed sources here as well. i.e. once a source is closed, new data is rejected. Feel free to grab sometime with Nitin (or Aditya) to understand how mock_client works.

nikhilgarg28:
Could you plz also document these in api-reference/sources/source-decorator? Don't worry too much about the fine language in concepts section - once this is pushed, I can take a final pass - just request you to create a Linear ticket for me.

docs/pages/concepts/source.md
saiharshavellanki: Updated the documentation. Let me know if it looks good

fennel/client_tests/test_bounded_idleness.py
```diff
@@ -0,0 +1,201 @@
...
+)
+@dataset
+class UnBoundedClicksDS:
+    display_id: int = field(key=True)
+    ad_id: int
+    clicked: bool
+    timestamp: datetime = field(timestamp=True)
+
+
+def _log_clicks_data_batch1(client, webhook_endpoint):
```
mohitreddy1996, line 42: Can you inline them? 

Looking at the test case, I have no idea what is getting logged so it is difficult to understand the expectations. For unit tests, we should definitely prioritize readability over code reusability.

fennel/client_tests/test_bounded_idleness.py
```diff
@@ -0,0 +1,201 @@
...
+
+    now = datetime.now()
+    ts = pd.Series([now, now, now])
+    display_id_keys = pd.Series([4, 5, 6])
+
+    df, _ = BoundedClicksDS.lookup(ts, display_id=display_id_keys)
+    assert df["ad_id"].to_list() == [5, 6, None]
+    assert df["clicked"].to_list() == [True, False, None]
+
+    # Sleep for 5s so that new data is not logged
```
mohitreddy1996, line 145: sleep for 10?

mohitreddy1996 (changes requested):
Looks good overall, requesting a small change in unit tests impl.

fennel/client_tests/test_bounded_idleness.py
saiharshavellanki: It is added for debugging purposes. Moved it back to 5s

fennel/client_tests/test_bounded_idleness.py
saiharshavellanki: Updated the logic to be more readable

mohitreddy1996 (approved):
@nikhilgarg28 can you take a look at the documentation changes? (I have not taken a pass at them)

nikhilgarg28:
I'd make changes myself latrer. @saiharshavellanki feel free to push it out for now.

PR #376, opened 2024-02-07
[DEV-2944] Added dataset versioning

nonibansal:


fennel/datasets/datasets.py
```diff
@@ -84,6 +85,7 @@
 
 DEFAULT_RETENTION = Duration("2y")
 DEFAULT_EXPIRATION = Duration("30d")
+DEFAULT_VERSION = 0
```
nikhilgarg28, line 88: I think it might be better to start versions from 1, not 0?

nonibansal: makes sense, will change I wrote 0 because I remember in featureset extractor we used to start from 0.

fennel/datasets/datasets.py
```diff
@@ -1031,27 +1039,42 @@ def f_get_type_hints(obj):
...
+    #     else:
+    #         callable_name = str(version)
+    #     raise ValueError(
+    #         f"pipeline decorator on `{callable_name}` must have a parenthesis"
+    #     )
+    # if type(version) is not int:
+    #     raise ValueError(
+    #         "pipeline version must be an integer, found %s" % type(version)
+    #     )
+
```
aditya-nambiar, line 1073: delete this code ?

fennel/datasets/datasets.py
```diff
@@ -1538,26 +1565,32 @@ def sync_validation_for_pipelines(pipelines: List[Pipeline], ds_name: str):
...
-            raise ValueError(
-                f"Multiple active pipelines are not supported for dataset {ds_name}."
-            )
-        if pipeline.active:
-            found_active = True
-    if not found_active and len(pipelines) > 1:
-        raise ValueError(
-            f"No active pipeline found for dataset {ds_name}. Please mark one of the pipelines as active by setting `active=True`."
-        )
+    if len(pipelines) > 1:
```
aditya-nambiar, line 1568: if len(pipelines) <= 1:
return

.. rest of code ..

This prevents indenting the whole code

aditya-nambiar (approved):

fennel/datasets/datasets.py
nonibansal: yup

fennel/datasets/datasets.py
nonibansal: changed this.

PR #377, opened 2024-02-08
docs: add detailed api reference for all sources

nikhilgarg28:


fennel/_vendor/README.md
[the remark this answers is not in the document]
nikhilgarg28: Done. Thanks!

fennel/datasets/datasets.py
```diff
@@ -1339,7 +1339,7 @@ def signature(self):
     def with_source(
         self,
         conn: DataConnector,
-        every: Optional[Duration] = None,
+        every: Duration = "1h",
```
nonibansal, line 1342: Any reason behind 1h, I feel like 1D should be a good default.

fennel/lib/to_proto/to_proto.py
```diff
@@ -691,8 +689,7 @@ def _kafka_to_ext_db_proto(
             sasl_plain_username=sasl_plain_username,
             sasl_plain_password=sasl_plain_password,
             sasl_jaas_config="",
-            # TODO(xiao): remove once server side is updated
-            enable_ssl_certificate_verification=BoolValue(value=verify_cert),
+            enable_ssl_certificate_verification=None,
```
nonibansal, line 692: Is this works? What if we pass false as default?

fennel/datasets/datasets.py
nonibansal: Oh alright.

PR #382, opened 2024-02-16
docs: add watermarking based metric documentation

mohitreddy1996:


nonibansal (approved):

docs/pages/observability/prometheus.md
```diff
@@ -90,6 +90,36 @@ The following metrics are exposed via the Prometheus endpoint:
    - `status` - whether expectation passed or not. Valid values are `success` or `failure`
    </details>
 </li>
+
+<li>
+<details>
+   <summary><b> `fennel_source_watermark_discarded_rows` </b></summary>
+
+   Reports the number of rows discarded at the source dataset due to watermarking i.e. the rows are older than the 
+   `disorderness` set on the source.
```
nikhilgarg28, line 99: There is no such thing as `disorderness` but only `disorder` (and if 
 that's not the case, we should fix it ASAP). @mohitreddy1996 could you plz fix this?

nikhilgarg28: Also, why not make this metric available for every dataset, not just sourced datasets?

nikhilgarg28 (commented):
Requesting some changes to be addressed in a follow up PR

docs/pages/observability/prometheus.md
mohitreddy1996: aah yes, late night PRs are not a good idea :/

mohitreddy1996: > Also, why not make this metric available for every dataset, not just sourced datasets?

others are implementation/system details? Not something the user can configure/use? And we should ideally never drop anything in the downstream systems right. So its an internal metric and not something for the user to worry/think about.

PR #389, opened 2024-02-26
[DEV 3365] Automatic addition of struct code in extractor

nonibansal:


aditya-nambiar (approved):
Did you test the client backend integration part of it too ?

PR #395, opened 2024-02-29
DEV-3190: Kafka connector params improvements

saiharshavellanki:


mohitreddy1996 (approved):
Do wait for @aditya-nambiar 's approval. He is more aware of the merge/release process on the client.

aditya-nambiar:
Please add the proto generations from the server side too

saiharshavellanki:
> Please add the proto generations from the server side too

Updated the generations

docs/pages/api-reference/sources/kafka.md
```diff
@@ -30,7 +30,7 @@ servers (e.g. `localhost:9092,another.host:9092`).
 Protocol used to communicate with the brokers. 
 </Expandable>
 
-<Expandable title="sasl_mechanism" type="Optional[str]">
+<Expandable title="sasl_mechanism" type="str">
 SASL mechanism (e.g. SCRAM-SHA-256, PLAIN) to use for authentication.
```
nikhilgarg28, line 34: Can you please enumerate the valid SASL mechanism values that we support?

nikhilgarg28: Also are `sasl_plain_username` and `sasl_plain_password` still optional?

nikhilgarg28 (changes requested):

docs/pages/api-reference/sources/kafka.md
saiharshavellanki: Added list of SASL mechanisms supported. Also made sasl_plain_username and sasl_plain_password as non-optional

fennel/sources/sources.py
```diff
@@ -255,7 +241,6 @@ def get(name: str) -> Kafka:
             sasl_mechanism="",
```
aditya-nambiar, line 241: qq why is securiy_protocol set to PLAINTEXT? I see sasl_mechanism as another literal being set to empty, should it have a default value too ?

aditya-nambiar (approved):

fennel/sources/sources.py
saiharshavellanki: As of now this method is being executed in test cases/documentation. That is the reason I haven't updated this logic. I believe specifying default values here doesn't change much given bootstrap servers are empty and there is no username and password.

PR #396, opened 2024-02-29
[DEV-3220] Add support for quantile aggregation

petrpan26:
Added support and mock client for quantile aggregation


fennel/datasets/aggregate.py
```diff
@@ -111,6 +111,7 @@ def signature(self):
 class Average(AggregateType):
     of: str
     default: float = 0.0
+    approx: bool = False
```
aditya-nambiar, line 114: Why is this added here ?

fennel/testing/execute_aggregation.py
```diff
@@ -337,6 +339,28 @@ def get_val(self) -> List:
...
+    def del_val_from_state(self, val):
+        self.vals.remove(val)
+        return self.get_val()
+
+    def get_val(self) -> List:
+        if len(self.vals) == 0:
+            return self.default
+        self.vals.sort()
+        index = int(math.floor(self.p * len(self.vals))) + 1
+        return self.vals[min(len(self.vals) - 1, max(index - 1, 0))]
```
aditya-nambiar, line 361: This would sort all the data for every read ?
Can we use a sorted tree data structure instead, inserts are log(n) and reads are log(n) too. 

from sortedcontainers import SortedList
class QuantileState(AggState):
    def __init__(self, default, p):
        self.p = p
        self.default = default
        self.vals = SortedList()

    def add_val_to_state(self, val):
        self.vals.add(val)
        return self.get_val()

    def del_val_from_state(self, val):
        self.vals.remove(val)
        return self.get_val()

    def get_val(self):
        if len(self.vals) == 0:
            return self.default
        index = int(math.floor(self.p * (len(self.vals) - 1)))
        return self.vals[index]

aditya-nambiar:
Can you add some tests for invalid Quantiles. 

Like Quantile over a string/lists, > 1 percentile, approx=False etc

Also a test for the generated proto.

aditya-nambiar:
Please add this to documentation too, and paste the screenshot in the description

fennel/datasets/aggregate.py
petrpan26: Must be a typo

petrpan26:
Screenshot 2024-03-13 at 12 30 15 AM
Screenshot 2024-03-13 at 12 30 27 AM
Screenshot 2024-03-13 at 12 30 31 AM

docs/examples/api-reference/aggregations/quantile.py
```diff
@@ -0,0 +1,322 @@
...
+            webhook = Webhook(name="webhook")
+
+            @source(
+                webhook.endpoint("Transaction"), disorder="14d", cdc="append"
+            )
+            @dataset
+            class Transaction:
+                uid: int
+                amount: int
+                # docsnip-highlight next-line
```
aditya-nambiar, line 298: why is this highlighted ?

docs/examples/api-reference/aggregations/quantile.py
```diff
@@ -0,0 +1,322 @@
...
```
aditya-nambiar, line 244: why is this high lighted ?

docs/examples/api-reference/aggregations/quantile.py
```diff
@@ -0,0 +1,322 @@
...
+
+                @pipeline
+                @inputs(Transaction)
+                def bad_pipeline(cls, ds: Dataset):
+                    return ds.groupby("uid").aggregate(
+                        Quantile(
+                            of="amount",
+                            window="1w",
+                            into_field="median_amount_1w",
+                            p=10.0,
```
aditya-nambiar, line 316: make < 1 ?

docs/examples/api-reference/aggregations/quantile.py
```diff
@@ -0,0 +1,322 @@
...
```
aditya-nambiar, line 262: highlight this ?

aditya-nambiar (approved):
Looks good to me. @nikhilgarg28 do take a pass

PR #403, opened 2024-03-02
Add more comprehensive error and documents for window and summary

petrpan26:
Add more comprehensive error and doc


aditya-nambiar:
Can you add screen shots of the documentation to the description

docs/pages/api-reference/operators/window.md
```diff
@@ -0,0 +1,60 @@
...
+Operator to do aggregation of events into windows based on their timestamps. Window operator must always be preceded by a [groupby](/api-reference/operators#groupby) operator.
+
+#### Parameters
+
+<Expandable title="type" type="Window">
+
+The type of window to group event on. Possible values are `"hopping"`, `"tumbling"`, `"session"`.
+
+`"hopping"`: Take in two parameters `"duration"` and `"stride"`, the `"duration"` parameters indicate how long the window is while `"gap"` indicate how far apart the window should be. Hopping window can overlap meaning one event can belong to two different windows. `"stride"` should be smaller than `"duration"`.
+
```
aditya-nambiar, line 19: You are specifying stide and duration but later using gap in the description ?

Can we change to something like -

A hopping window is defined by two key parameters: duration and stride.

< sublist >
duration: The length of the window, indicating the time span that each window covers.

stride (also referred to as gap in some contexts): The interval between the start of each window. This setting allows for the creation of overlapping windows, where a single event can be included in multiple windows. It's essential that the stride is less than the duration.

docs/pages/api-reference/operators/window.md
```diff
@@ -0,0 +1,60 @@
...
+`"tumbling"`: Take in one parameter `"duration"` indicate how long the window is. The tumbling windows do not overlapped and cover the whole range.
```
aditya-nambiar, line 20: Tumbling windows are characterized by a single parameter: duration.

duration: Specifies how long each window lasts. Unlike hopping windows, tumbling windows are designed to be non-overlapping and contiguous, ensuring that each event is placed into exactly one window, covering the entire data range without gaps.

docs/pages/api-reference/operators/window.md
```diff
@@ -0,0 +1,60 @@
...
+
+<Expandable title="type" type="Window">
+
+The type of window to group event on. Possible values are `"hopping"`, `"tumbling"`, `"session"`.
+
+`"hopping"`: Take in two parameters `"duration"` and `"stride"`, the `"duration"` parameters indicate how long the window is while `"gap"` indicate how far apart the window should be. Hopping window can overlap meaning one event can belong to two different windows. `"stride"` should be smaller than `"duration"`.
+
+`"tumbling"`: Take in one parameter `"duration"` indicate how long the window is. The tumbling windows do not overlapped and cover the whole range. 
+
+`"session`: Take in one parameter `"gap"` indicating how close together two events are to be group into one window. Session window do not overlapped and always at least `"gap"` apart.
```
aditya-nambiar, line 22: Session windows are dynamically sized based on event patterns, using the gap parameter.

gap: Determines the timeout for inactivity that defines the boundary of a window. Events that occur closer together than the specified gap are grouped into the same window. This approach allows windows to be created based on the actual occurrence of events, with windows being at least gap duration apart.

docs/examples/api-reference/operators/summarize.py
```diff
@@ -0,0 +1,119 @@
...
+            @inputs(Transaction)
+            def window_pipeline(cls, ds: Dataset):
+                # docsnip-highlight start
+                return (
+                    ds.groupby("uid")
+                    .window(type="session", gap="15m", field="window")
+                    .summarize(
+                        field="total_amount",
+                        dtype=int,
+                        func=lambda df: df["amount"].sum(),
```
nikhilgarg28, line 54: Why are these taking kwarg and not positional args like `assign` operator?

docs/examples/api-reference/operators/summarize.py
```diff
@@ -0,0 +1,119 @@
...
+            # docsnip-highlight end
+            timestamp: datetime = field(timestamp=True)
+
+            @pipeline
+            @inputs(Transaction)
+            def window_pipeline(cls, ds: Dataset):
+                # docsnip-highlight start
+                return (
+                    ds.groupby("uid")
+                    .window(type="session", gap="15m", field="window")
```
nikhilgarg28, line 50: I don't think `field` as kwarg is clear enough - could we change this to make it more human readable in english like a sentence? If nothing else, we should use `into_field` since that's what we use for aggregations (though i don't like even that).

docs/examples/api-reference/operators/summarize.py
```diff
@@ -0,0 +1,119 @@
...
+            uid: int
+            amount: int
+            timestamp: datetime = field(timestamp=True)
+
+        @dataset
+        class Session:
+            # docsnip-highlight start
+            # groupby field becomes the key field
+            uid: int = field(key=True)
+            # new fields are added to the dataset by the window operation
```
nikhilgarg28, line 38: Mention that window also becomes a key

docs/pages/api-reference/operators/window.md
```diff
@@ -0,0 +1,60 @@
...
+---
+### Window
+
+<Divider>
+<LeftSection>
+Operator to do aggregation of events into windows based on their timestamps. Window operator must always be preceded by a [groupby](/api-reference/operators#groupby) operator.
+
+#### Parameters
+
+<Expandable title="type" type="Window">
```
nikhilgarg28, line 14: Here you should say `type="hopping" | "tumbling" | "session".

docs/pages/api-reference/operators/window.md
```diff
@@ -0,0 +1,60 @@
...
+
+The type of window to group event on. Possible values are `"hopping"`, `"tumbling"`, `"session"`.
+
+`"hopping"`: Take in two parameters `"duration"` and `"stride"`, the `"duration"` parameters indicate how long the window is while `"gap"` indicate how far apart the window should be. Hopping window can overlap meaning one event can belong to two different windows. `"stride"` should be smaller than `"duration"`.
```
nikhilgarg28, line 18: This is double documentation - would instead suggest to mention something simpler here e.g. "If window is "hopping", `duration` and `stride` must be set. ....

And then do the acutal documentation of all these params in their own expandable section.

docs/pages/api-reference/operators/window.md
```diff
@@ -0,0 +1,60 @@
...
+<Expandable title="stride" type="Duration">
+
+The parameters for `"hopping"` windows. Indicating how far the windows from each other. Possible values are any [time duration](/api-reference/data-types/duration)
+</Expandable>
+
+<Expandable title="field" type="str">
+The name of the field in the output dataset that should store the result of this aggregation. This field is expected to be of type `"Window"`.
+</Expandable>
+
+#### Returns
```
nikhilgarg28, line 44: Can you plz document the `Errors` section as well - what are most important commit type or runtime errors?

nikhilgarg28 (changes requested):
Requesting changes.

docs/examples/api-reference/operators/summarize.py
nikhilgarg28: btw I don't feel as strongly about this as I do about `summarize` feeling similar to `assign` in how it's written.

petrpan26: I think this is reasonable !

petrpan26:
Screenshot 2024-03-13 at 2 19 38 AM
Screenshot 2024-03-13 at 2 19 45 AM
Screenshot 2024-03-13 at 2 19 52 AM
Screenshot 2024-03-13 at 2 19 57 AM
Screenshot 2024-03-13 at 2 20 02 AM
Screenshot 2024-03-13 at 2 20 09 AM

aditya-nambiar (approved):

PR #411, opened 2024-03-08
err: Ensure sync errors are displayed correctly

aditya-nambiar:


fennel/client/client.py
```diff
@@ -779,3 +776,15 @@ def _s3_connector_dict(s3: S3Connector) -> Dict[str, Any]:
         "db": {"S3": {"creds": creds_json}},
     }
     return s3_table
+
+
+def _validate_branch_name(branch: str):
+    """
+    Branch name should only contain alphanumeric characters, hyphens or underscores.
```
nikhilgarg28, line 783: Why are we adding this restriction? I think it's important to support "/" since many branches will be of the form "author/something". I think we should stick with the equivalent of git branch names: https://stackoverflow.com/questions/3651860/which-characters-are-illegal-within-a-branch-name

fennel/client/client.py
```diff
@@ -311,6 +306,8 @@ def init_branch(self, name: str):
         Raises an exception if a branch with the same name already exists.
 
         """
+        print(f"Creating branch {name}...")
```
nikhilgarg28, line 309: Stray?

aditya-nambiar: I felt its useful, to know,  since sometimes you can have invalid names etc, now its part of the log

fennel/client/client.py
aditya-nambiar: Yeah that is precisely why I have added this restriction. author/name wont work , since it can break at a bunch of places, like we pass it in the URLs etc

aditya-nambiar: Currently the directive I planning is - <author_name>-<branch_name>

fennel/client/client.py
nikhilgarg28: Upto you. In my view, we haven't put these prints anywhere else. And I think the real issue is that print is not a reliable communication mechanism. If there was any information that we wanted user to know, it should either be in return value or exception.

PR #418, opened 2024-03-17
Protect against aloglia duplicating records

LukeSmetham:
Adds a regex pattern to `stop_urls` in the algolia config so that we don't get duplicated records when the scraper follows `<a>` tags.

This was happening for two reasons: 

First was the originally we we're adding all possible urls in the API reference to the sitemap - which is irrelevant as all `/api-reference/*` pages have identical content (this is now fixed in `turbo`) - so we would get the same content scraped for every possible path

Second was that algolias scraper treats each url in the sitemap as a `start_url` and will follow any `<a>` tags like the google search spider. The issue here is that internal links between api-reference pages would have the same effect as above, where the same page is being scraped multiple times

With this PR we're telling Aloglia to only allow `/api-reference` and `/api-reference/x.x.x` and to ignore any urls deeper than this


LukeSmetham:
I ran the scraper image locally so search is now fixed on the docs, but we need this landed in order to maintain this functionality on subsequent content updates

nikhilgarg28 (approved):

PR #421, opened 2024-03-19
Adding encoding of branch name in network calls

nonibansal:


aditya-nambiar (approved):
Please increment version and add to changelog

aditya-nambiar:
Can we also add to docs on how to make a rest call, with an example ?

PR #425, opened 2024-03-23
[DEV 2914] Removed id and provider from feature

nonibansal:
https://linear.app/fennel-ai/issue/DEV-2916/remove-provider-from-auto-generated-features
https://linear.app/fennel-ai/issue/DEV-2914/deprecate-featureids=1

## Summary by CodeRabbit

- **Documentation**
	- Updated API references, concepts, datasets, examples, and testing documentation to reflect changes in feature handling and extraction syntax.
	- Enhanced clarity and consistency in feature declaration and usage across various examples and test cases.
- **Refactor**
	- Simplified feature import statements across multiple files.
	- Removed explicit `id` assignments in feature declarations, aligning with updated feature handling practices.
	- Adjusted feature extraction syntax and annotations for improved clarity and consistency.
- **New Features**
	- Introduced string literals for `@inputs` and `@outputs` decorators to enhance code readability and maintainability.
	- Added `@index` decorator to support indexing on specific class attributes.
- **Bug Fixes**
	- Fixed issues related to feature extraction and declaration inconsistencies across documentation and test files.
- **Tests**
	- Updated unit tests to align with the new feature declaration and extraction practices.


docs/examples/api-reference/client/query.py
```diff
@@ -14,14 +14,14 @@
...
     from fennel.lib import inputs, outputs
 
     @featureset
     class Numbers:
-        num: int = feature(id=1)
-        is_even: bool = feature(id=2)
-        is_odd: bool = feature(id=3)
+        num: int = F()
+        is_even: bool = F()
+        is_odd: bool = F()
```
nikhilgarg28, line 24: Just to confirm, we don't need to write `F()` here, right? As in, when there is no qualifier, we can simply omit this like we do with fields? cc @aditya-nambiar

docs/examples/api-reference/client/commit.py
```diff
@@ -36,9 +36,9 @@ class Transaction:
 
     @featureset
     class TransactionFeatures:
-        txid: int = feature(id=1)
-        amount: int = feature(id=2).extract(field=Transaction.amount, default=0)
-        amount_is_high: bool = feature(id=3)
+        txid: int = F()
+        amount: int = F(ref=Transaction.amount, default=0)
```
nikhilgarg28, line 40: I think the conclusion was to not use any kwarg but the first position arg to be the expression? So no `ref` here?

nikhilgarg28 (changes requested):

docs/examples/api-reference/client/query.py
aditya-nambiar: We are currently running into an issue where if the feature is not initialized with anything, we are unable to use it in an extractor ( as inputs or outputs )

nikhilgarg28: Do we have any viable paths of resolving this? Or do you expect to have to live with this at this stage?

docs/examples/api-reference/client/commit.py
nonibansal: will replace this

docs/examples/api-reference/client/query.py
nonibansal: One option was to make inputs and outputs take string like @inputs("feature1") but this makes the code look bad.

nonibansal: moving with combination then as discussed.

aditya-nambiar (approved):

PR #429, opened 2024-04-01
Add client code + doc for kafka sink

petrpan26:
Added sinks client API for Kafka and added documentation.


petrpan26:
Screenshot 2024-04-01 at 6 03 04 PM
Screenshot 2024-04-01 at 6 03 10 PM
Screenshot 2024-04-01 at 6 03 14 PM
Screenshot 2024-04-01 at 6 03 27 PM

fennel/_vendor/pydantic/env_settings.py
```diff
@@ -55,7 +55,7 @@ def _build_values(
         _env_nested_delimiter: Optional[str] = None,
         _secrets_dir: Optional[StrPath] = None,
     ) -> Dict[str, Any]:
-        # Configure built-in sources
+        # Configure built-in connectors
```
mohitreddy1996, line 58: looks like you made a replace all "sources" -> "connectors" :P 

These changes are in pydantic which are not relevant to this wider change

fennel/connectors/__init__.py
```diff
@@ -1,3 +1,4 @@
+# for backward compatibility, will be removed in the future.
```
mohitreddy1996, line 1: what is the context here? AFAIU, this is a stale comment?

fennel/connectors/connector.py
mohitreddy1996, line 1: I could be totally wrong, but this looks like a file copy of connectors.py in the same module. Was there a rebasing issue or this is intended?

mohitreddy1996 (changes requested):

docs/examples/api-reference/sinks/kafka_sinks.py
```diff
@@ -0,0 +1,69 @@
...
+
+@mock
+def test_kafka_sink(client):
+    os.environ["KAFKA_USERNAME"] = "test"
+    os.environ["KAFKA_PASSWORD"] = "test"
+
+    from fennel.connectors import source, Kafka
+    from fennel.datasets import dataset, field
+
+    kafka = Kafka(
```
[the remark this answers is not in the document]
petrpan26: Makes sense ! Move this doc snip down below

docs/examples/api-reference/client/erase.py
```diff
@@ -12,7 +12,7 @@ def test_basic(client):
     from fennel.datasets import dataset, field, index
     from fennel.connectors import source, Webhook
 
-    # first define & sync a dataset that sources from a webhook
+    # first define & sync a dataset that connectors from a webhook
```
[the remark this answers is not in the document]
petrpan26: This is a refactor error

docs/examples/api-reference/sources/sink.py
```diff
@@ -0,0 +1,49 @@
+import os
+from datetime import datetime
+
+from fennel.testing import mock
+
+__owner__ = "nikhil@fennel.ai"
+
+
+@mock
+def test_sink_decorator(client):
```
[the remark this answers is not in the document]
petrpan26: I think i was supposed to delete this test

fennel/internal_lib/to_proto/to_proto.py
```diff
@@ -643,6 +688,19 @@ def _conn_to_source_proto(
         raise ValueError(f"Unknown connector type: {type(connector)}")
 
 
+def _conn_to_sink_proto(
+    connector: connectors.DataConnector,
+    dataset_name: str,
+    ds_version: int,
+) -> Tuple[connector_proto.ExtDatabase, connector_proto.Source]:
```
mohitreddy1996, line 695: connector_proto.Sink and not source I would imagine?

fennel/connectors/test_invalid_connectors.py
mohitreddy1996, line 1: same as above, is this a copy of test_invalid_sources?

mohitreddy1996 (changes requested):

fennel/connectors/test_invalid_connectors.py
petrpan26: Removed

fennel/connectors/connectors.py
```diff
@@ -112,17 +112,25 @@ def decorator(dataset_cls: T):
 
 
 def sink(
-    conn: DataConnector, every: Optional[Duration] = None
+    conn: DataConnector,
+    cdc: str,
+    tier: Optional[Union[str, List[str]]] = None,
 ) -> Callable[[T], Any]:
+    if not isinstance(conn, DataConnector):
```
mohitreddy1996, line 119: Let's add restrictions on what do we allow as of today to be defined as sinks? 

We should ideally do that so that the user gets errors as part of "commit" without communication with the backend at all. This will allow the code paths to be tested even with mock clients which is the default local development workflow of our users

fennel/connectors/test_connectors.py
```diff
@@ -508,10 +540,38 @@ class UserInfoDataset:
...
+    assert sink_request == expected_sink_request, error_message(
+        sink_request, expected_sink_request
+    )
     sync_request = view._get_sync_request_proto(tier="staging")
-    assert len(sync_request.datasets) == 1
+    assert len(sync_request.datasets) == 2
     assert len(sync_request.sources) == 1
-    assert len(sync_request.extdbs) == 1
+    assert len(sync_request.sinks) == 1
+    assert len(sync_request.extdbs) == 2
```
mohitreddy1996, line 574: so extdbs here is 2 because we have 1 source (snowflake) for tier = staging and now we have defined a sink (kafka). Can you also ensure that if the same kafka cluster is used in both source and sink, we create a single extdb?

fennel/internal_lib/to_proto/to_proto.py
```diff
@@ -134,17 +135,28 @@ def to_sync_request_proto(
...
-                continue
-            (ext_db, s) = res
-            conn_sources.append(s)
-            # dedup external dbs by the name
-            # TODO(mohit): Also validate that if the name is the same, there should
-            # be no difference in the other fields
-            if ext_db.name not in external_dbs_by_name:
-                external_dbs_by_name[ext_db.name] = ext_db
-                external_dbs.append(ext_db)
+            sources = sources_from_ds(obj, obj.timestamp_field, tier)
```
mohitreddy1996, line 138: can we make sources_from_ds identical to sinks_from_ds? i.e. not return an Optional[] but actually list as you have done. It would make code flow a bit easier to follow (e.g. how it is for sinks right now)

Any reason why these two have different return types (they are extracted using similar information and the only diff is the source/sink).

fennel/connectors/connectors.py
mohitreddy1996: I see that we have restriction on Kafka Connector as the sink at the proto conversion layer, but may be we should make it more explicit at the sink creator/decorator itself? Also we need to ensure the cdc conversion because proto conversion code allows literally any cdc method as of right now

mohitreddy1996 (changes requested):
Looking good, final set of comments

fennel/internal_lib/to_proto/to_proto.py
mohitreddy1996: Understood

mohitreddy1996 (approved):
beautiful1

PR #446, opened 2024-04-15
[DEV 3854] Adding decimal type

nonibansal:


fennel/datasets/datasets.py
```diff
@@ -111,7 +112,7 @@
 DEFAULT_INDEX_OFFLINE = "forever"
 DEFAULT_INDEX_ONLINE = True
 
-primitive_numeric_types = [int, float, pd.Int64Dtype, pd.Float64Dtype]
+primitive_numeric_types = [int, float, pd.Int64Dtype, pd.Float64Dtype, Decimal]
```
aditya-nambiar, line 115: all this will go with pandas 2.0 ?

aditya-nambiar (approved):

fennel/datasets/datasets.py
nonibansal: yup this will work with Pandas 2.0, basically this is for user provided dtype in the Dataset class.

PR #448, opened 2024-04-16
Support aggregation along on another timestamp

petrpan26:
Added support for client code and mock client to support aggregation on another time axis.


petrpan26:
Documentation change
Screenshot 2024-04-16 at 12 11 47 AM
Screenshot 2024-04-16 at 12 11 52 AM

fennel/datasets/datasets.py
```diff
@@ -507,10 +519,19 @@ def aggregate(self, *args, **kwargs) -> _Node:
         if len(self.keys) == 1 and isinstance(self.keys[0], list):
             self.keys = self.keys[0]  # type: ignore
 
+        if along is not None and not isinstance(along, str):
+            raise ValueError(
+                f"Along argument have to be string or None got: {along}"
```
nikhilgarg28, line 524: improve the error:

`along kwarg in the aggregate operator must be string or None, found: `{along}``

fennel/datasets/datasets.py
```diff
@@ -2298,6 +2319,21 @@ def visitAggregate(self, obj) -> DSSchema:
...
         values: Dict[str, Type] = {}
+        if isinstance(obj.along, str):
+            along_type = input_schema.values.get(obj.along)
+            if along_type is None:
+                raise ValueError(
+                    f"Along field `{obj.along}` does not exist in input schema non-key non-timestamp values."
+                )
+            if along_type != datetime.datetime:
+                raise ValueError(
+                    f"Along field `{obj.along}` must be a `timestamp` field."
```
nikhilgarg28, line 2330: `error with along kwarg of aggregate operator: not of type datetime` (note that we don't need this to be the timestamp field, just any datetime field)

fennel/datasets/datasets.py
```diff
@@ -2298,6 +2319,21 @@ def visitAggregate(self, obj) -> DSSchema:
...
+                )
+        else:
+            if obj.along is not None:
+                raise ValueError(
+                    f"Expected along field `{obj.along}` to be str field"
```
nikhilgarg28, line 2335: Similarly improve this error.

fennel/datasets/datasets.py
```diff
@@ -2394,10 +2430,14 @@ def visitAggregate(self, obj) -> DSSchema:
...
                 raise TypeError(f"Unknown aggregate type {type(agg)}")
+
+        timestamp_field = (
+            obj.along if obj.along is not None else input_schema.timestamp
+        )
         return DSSchema(
             keys=keys,
             values=values,  # type: ignore
-            timestamp=input_schema.timestamp,
+            timestamp=timestamp_field,
```
nikhilgarg28, line 2440: Does this mean that the name of the along field becomes the name of the timestamp field in output dsschema? If so, that's not the desired behavior. The timestamp field of the input dataset should still be the name of the timestamp field in the output dataset - just that lookup will behave as if it was along a different axis.

fennel/client_tests/test_dataset.py
```diff
@@ -1206,6 +1285,66 @@ def test_basic_aggregate(self, client):
...
+        ]
+        columns = ["userid", "rating", "movie", "t", "rating_time"]
+        df = pd.DataFrame(data, columns=columns)
+        response = client.log("fennel_webhook", "RatingActivity", df)
+        assert response.status_code == requests.codes.OK
+
+        client.sleep()
+
+        # Do some lookups to verify pipeline_aggregate is working as expected
+        ts = pd.Series([now, now])
```
nikhilgarg28, line 1323: Could you please test this on some other timestamps too such that more complex interaction of along and timestamp are tested?

nikhilgarg28 (changes requested):

fennel/client_tests/test_dataset.py
petrpan26: Will add more test query back in time but still following the hacky along like server side right now

nikhilgarg28 (approved):

PR #458, opened 2024-04-24
Support Google Pub/Sub source

saiharshavellanki:
Added support for Google Pub/Sub data source
Updated the documentation

Screenshot 2024-04-24 at 12 07 25 PM

Screenshot 2024-04-24 at 12 07 38 PM

Screenshot 2024-04-24 at 12 07 50 PM


docs/examples/api-reference/sources/sql.py
```diff
@@ -218,3 +218,40 @@ class UserClick:
...
+            "client_id": "103688493243243272951",
+            "auth_uri": "https://accounts.google.com/o/oauth2/auth",
+            "token_uri": "https://oauth2.googleapis.com/token",
+            "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
+        },
+    )
+    # docsnip-highlight end
+
+    # docsnip-highlight next-line
+    collection = pubsub.topic("test_topic")
```
nikhilgarg28, line 247: Rename `collection` to `topic`. In fact, try to inline `pubsub.topic(...)` in the `@source` decorator in a single line - if you can, do that. Else, keep it like this across two lines. You may have to shorten `test_topic` to say `name`

docs/pages/api-reference/connectors/pubsub.md
```diff
@@ -0,0 +1,70 @@
...
+A name to identify the source. The name should be unique across all Fennel sources.
+</Expandable>
+
+<Expandable title="project_id" type="str">
+The project ID of the Google Cloud project containing the Pub/Sub topic
+</Expandable>
+
+<Expandable title="service_account_key" type="str">
+The JSON string containing the credentials for the Service Account to use to access
+Pub/Sub topic. See below for instructions on how to obtain this.
```
nikhilgarg28, line 20: This isn't a json string but a dictionary, right? If so, please change this line to mention that.

docs/pages/api-reference/connectors/pubsub.md
```diff
@@ -0,0 +1,70 @@
...
+</Expandable>
+
+:::info
+Fennel supports only Upsert mode CDC with data in JSON format. If you require support
+for schema or CDC data format, please reach out to Fennel support.
+:::
```
nikhilgarg28, line 26: I don't follow this. What are we trying to say? Can use not send debezium data in pub-sub? Similarly, can use not do `append` mode only for keyless datasets?

docs/pages/api-reference/connectors/pubsub.md
```diff
@@ -0,0 +1,70 @@
...
+
+#### Topic Parameters
+<Expandable title="topic_id" type="str">
```
nikhilgarg28, line 29: Is `topic_id` a standard GCP thing? If not, let's just call it `name` since `id` usually denotes an integer, not a string.

docs/pages/api-reference/connectors/pubsub.md
```diff
@@ -0,0 +1,70 @@
...
+Then, add the service account as a Member of your Google Cloud Project with the "Pub/Sub Subscriber" role. To do this, follow
+the instructions
+for [Granting Access](https://cloud.google.com/iam/docs/granting-changing-revoking-access#granting-console) in the
+Google documentation. The email address of the member you are adding is the same as the Service Account ID you just
+created.
+
+At this point, you should have a service account with the "Pub/Sub Subscriber" project-level permission.
+
+For Service Account Key JSON, enter the Google
+Cloud [Service Account Key in JSON format](https://cloud.google.com/iam/docs/creating-managing-service-account-keys).
```
nikhilgarg28, line 66: This part isn't clear and needs to be reworded. I suppose you mean to simply say something like "To obtain a Service Account Key, follow the instructions on <link to page>" or something like that?

nikhilgarg28 (changes requested):

docs/pages/api-reference/connectors/pubsub.md
mohitreddy1996: yeah its a pub-sub specific thing -

docs/pages/api-reference/connectors/pubsub.md
mohitreddy1996: @nikhilgarg28 there are some limitations with Pubsub inherently around ordered delivery of messages which prevents us from performing exactly-once semantics. We have listed down some of it here - https://coda.io/d/_dmX2EX1n2_s/Pub-Sub-Fennel-Documentation_suRHk

mohitreddy1996 (changes requested):
@saiharshavellanki I think we should anyways add format as an input to the pub sub topic decorator. We can add validations for now that JSON is the only valid format for now, but let's setup the base for potential highly likely extensions.

docs/pages/api-reference/connectors/pubsub.md
saiharshavellanki: That is correct. Additionally, I am rewording to "supports only Append and Upsert mode with CDC" since we can support Append as well similar to other

docs/pages/api-reference/connectors/pubsub.md
saiharshavellanki: Yes that is the standard naming convention

docs/pages/api-reference/connectors/pubsub.md
saiharshavellanki: That is correct. Let me reword it

docs/pages/api-reference/connectors/pubsub.md
saiharshavellanki: I copied this from BigQuery and forgot to update. Updated now

saiharshavellanki:
> @saiharshavellanki I think we should anyways add format as an input to the pub sub topic decorator. We can add validations for now that JSON is the only valid format for now, but let's setup the base for potential highly likely extensions.

Added format as input to decorator and added validation

mohitreddy1996 (approved):

PR #460, opened 2024-04-25
ci: Use depot action runners

ba1ajinaidu:


jacobwgillespie:
Hey @ba1ajinaidu @mohitreddy1996 it looks like these jobs are queued because of a permissions issue with the default runner group in your org, so GitHub isn't releasing them to the Depot runners: https://github.com/organizations/fennel-ai/settings/actions/runner-groups/1. 

You may want to check that page to see if the "allow public repositories" checkbox is checked:

image

PR #461, opened 2024-04-29
[DEV 4043] Adding fennel secrets

nonibansal:


fennel/client/client.py
```diff
@@ -759,6 +759,34 @@ def inspect(
 
         return self._parse_dataframe(result, output_dtypes)
 
+    # ----------------------- Secret API's -----------------------------------
+
+    def get_secret(self, secret_name: str) -> Optional[str]:
```
aditya-nambiar, line 764: I set a set_secrets in the mock client but not here? lets keep parity and add it to both ?

fennel/client/client.py
```diff
@@ -759,6 +759,34 @@ def inspect(
...
+        data = {"value": secret_value}
+        response = self._post_json(
+            "{}/secret/{}".format(V1_API, secret_name),
+            data,
+        )
+        if response.status_code != requests.codes.OK:
+            raise Exception(response.json())
+        return response
+
+    def update_secret(self, secret_name: str, secret_value: str):
```
aditya-nambiar, line 780: lets add a delete secret too ?

fennel/client/client.py
```diff
@@ -844,6 +881,29 @@ def _post(
...
         return response
 
+    def _patch(
+        self,
+        path: str,
+        data: Any,
+        headers: Dict[str, str],
+        compress: bool = False,
+        timeout: float = _DEFAULT_TIMEOUT,
+    ):
```
aditya-nambiar, line 891: can you add a description

fennel/client/client.py
```diff
@@ -844,6 +881,29 @@ def _post(
...
+        if compress:
+            data = gzip.compress(data)
+            headers["Content-Encoding"] = "gzip"
+        if self.token:
+            headers["Authorization"] = "Bearer " + self.token
+        response = self.http.request(
+            "PATCH",
```
aditya-nambiar, line 898: set secrets would be PUT though right

aditya-nambiar (approved):

fennel/client/client.py
nonibansal: set should be through PUT but makes sense to remove set, to maintain consistent APIs.

fennel/client/client.py
nonibansal: Abhay suggested to have secrets immutable, should we allow users to delete secrets?

fennel/client/client.py
nonibansal: yup, reomved set_secret in mock client.

fennel/client/client.py
nonibansal: added

fennel/client/client.py
nonibansal: Added the capability, basically soft deleted the mapping to execute the same.

PR #462, opened 2024-04-30
[DEV 4014] Support changing version of dataset in incremental mode in mock client

nonibansal:


fennel/client_tests/test_incremental.py
```diff
@@ -0,0 +1,67 @@
...
+from fennel.datasets import dataset, field
+from fennel.testing import mock
+
+__owner__ = "nitin@fennel.ai"
+webhook = Webhook(name="fennel_webhook")
+
+
+@pytest.mark.integration
+@mock
+def test_dataset_version_change(client):
```
aditya-nambiar, line 15: can you modify the test to change something and actually assert the change took place.
for eg you had a pipeline with transform doing x2 and later x3, on version increment with incremental mode we should see x3

fennel/testing/mock_client.py
```diff
@@ -134,14 +134,45 @@ def is_superset_featureset(featureset_new, featureset_old):
...
+            is the new dataset eligible or not. Eligible if:
+            1. If version same and old == new then eligible.
+            2. new != old then version should be higher.
+            """
+            if dataset_new.version > dataset_old.version:
+                return True
+            elif dataset_new.version < dataset_old.version:
+                return False
+            else:
+                return dataset_old.signature() == dataset_new.signature()
```
aditya-nambiar, line 149: we were planning to deprecate signatures earlier, is there another check we can use or maybe define an equality instead of using signature ?

fennel/client_tests/test_incremental.py
```diff
@@ -0,0 +1,67 @@
...
+    # Try changing the dataset with increasing the version should pass
+    @source(
+        webhook.endpoint("Dataset1"),
+        disorder="14d",
+        cdc="upsert",
+    )
+    @dataset(version=2)
+    class Dataset1:
+        user_id: int = field(key=True)
+        email: str
```
aditya-nambiar, line 62: can you also add a test where we incrementally change an extractor logic, to test out the read side

fennel/client_tests/test_incremental.py
nonibansal: changed

fennel/client_tests/test_incremental.py
nonibansal: added

fennel/testing/mock_client.py
nonibansal: would need equality at multiple places for the same, ie datasets, pipelines. Can take this task as another linear ticket.

aditya-nambiar (approved):

PR #463, opened 2024-05-06
[DEV 4091] Added date type

nonibansal:


aditya-nambiar (approved):

aditya-nambiar:
what errro do we throw if i set field(timestamp=True) for a date field

nonibansal:
> what errro do we throw if i set field(timestamp=True) for a date field

I can add a specific error but how about, only datetime dtypes can marked as timestamp field?

PR #467, opened 2024-05-13
data connector: Support headless csv files and arbitrary delimiter

aditya-nambiar:


fennel/connectors/connectors.py
```diff
@@ -224,6 +224,7 @@ def bucket(
         format: str = "csv",
         presorted: bool = False,
         spread: Optional[Duration] = None,
+        headers: Optional[List[str]] = None,
```
nikhilgarg28, line 227: One option to consider - don't put delimitter/headers etc in s3 but instead create a class for CSV format specifier and put these as kwargs in that. With that, people could write format='csv' or format=CSV(delimiter='\t\, headers=...). With this, CSV format could be used in other relevant places too and s3 source doesn't become too long. Similar pattern will be useful for other kinds of formats as well (e.g. imagine avro registry going inside the avro format class - which is probably what we already do).

mohitreddy1996: +1 - given headers and delimiter are very specific to format, we should probably make a type for it and try to colocate things as much as possible

aditya-nambiar: I had a similar question in the backend too, and chose to keep it in S3, since FileFormat is a nice clean struct I dint want to mess around by adding headers which doesn't fit in too well. 

Maybe here it isn't too bad

aditya-nambiar: actually format is kwarg is going to be used WAY more frequently then headers, and making it of type Optional[str, CSV] is not right. 

Opting to keep it this way in favor of keeping the common case cleaner

bothra90 (approved):

PR #474, opened 2024-05-17
[DEV 4161] Support nano seconds in timestamp field in mock client

nonibansal:


aditya-nambiar:
server is all microseconds, i was hoping to cast nano to micros and log a warning

PR #479, opened 2024-05-23
[FIX] Fix null check in pandas dtype conversion

nonibansal:


nikhilgarg28 (commented):
Can you plz write a unit test where this bug fix would have given wrong answers earlier?

nonibansal:
> Can you plz write a unit test where this bug fix would have given wrong answers earlier?

yup will write one.

aditya-nambiar (approved):

PR #490, opened 2024-06-12
DEV-4284: Support credential based authentication for Redshift

saiharshavellanki:
- Add support for credential based authentication for Redshift
- Updated the documentation

Screenshot 2024-06-12 at 3 37 38 PM

Screenshot 2024-06-12 at 3 37 44 PM


nikhilgarg28:
"Do not set this parameter in case we are going with IAM authentication" -> "Do not set this parameter when using IAM authentication"

"Do not set this parameter in case we are going with credential based authentication" -> "Do not set this parameter when using username/password for authentication"

aditya-nambiar (approved):

PR #492, opened 2024-06-17
[DEV 4304] Add capability to provide featureset as str in query/query_offline

nonibansal:


fennel/testing/mock_client.py
```diff
@@ -518,6 +524,33 @@ def _get_secret(self, secret_name: str) -> Optional[str]:
...
+    ) -> List[Union[Feature, Featureset, str]]:
+        output_feature_names: List[Union[Feature, Featureset, str]] = []
+        for output in outputs:
+            if isinstance(output, Feature):
+                output_feature_names.append(output)
+            elif isinstance(output, Featureset):
+                output_feature_names.append(output)
+            elif isinstance(output, str):
+                try:
+                    is_valid_feature(output)
```
aditya-nambiar, line 539: why is is_valid_feature throwing an exception rather than returning a bool ?

nonibansal: changed to bool.

aditya-nambiar (approved):

PR #501, opened 2024-06-29
Protobuf Format Support for Kafka Connector

saiharshavellanki:
- Added support for Protobuf format for Kafka connector
- Updated the documentation 

Screenshot 2024-06-29 at 8 13 00 AM
Screenshot 2024-06-29 at 8 13 09 AM

**Kafka topic parameters**
Screenshot 2024-06-29 at 8 14 21 AM


docs/pages/api-reference/connectors/kafka.md
```diff
@@ -44,9 +44,9 @@ SASL password.
 The name of the kafka topic that needs to be sourced into the dataset.
 </Expandable>
 
-<Expandable title="format" type='"json" | Avro' defaultVal="json">
-The format of the data in Kafka topic. Both `"json"` and 
-[Avro](/api-reference/connectors/avro) supported.
+<Expandable title="format" type='"json" | Avro | Protobuf' defaultVal="json">
+The format of the data in Kafka topic. `"json"`, 
+[Avro](/api-reference/connectors/avro) and [Protobuf](/api-reference/connectors/protobuf) supported.
```
nikhilgarg28, line 49: Does protobuf already work with Kinesis too or does it need a bit more work? If it already works, should we update similar documentation for Kinesis too?

docs/pages/api-reference/connectors/protobuf.md
```diff
@@ -0,0 +1,46 @@
...
+</Expandable>
+
+<Expandable title="token" type="Optional[str]">
+Token to be used for authentication with the schema registry. Only one of 
+username/password or token must be provided.
+</Expandable>
+
+<pre snippet="api-reference/sources/kafka#kafka_with_protobuf"
+    status="success" message="Using protobuf registry with kafka"
+    highlight="13-18, 20">
```
nikhilgarg28, line 45: Request to not use this but instead use `docsnip-highlight <something>` tags inline with the code. That way, as code gets refactored, the highlight stays correct even if actual line numbers change.

fennel/connectors/connectors.py
```diff
@@ -347,6 +347,14 @@ class Avro(BaseModel):
     token: Optional[str]
 
 
+class Protobuf(BaseModel):
+    registry: str
+    url: str
+    username: Optional[str]
+    password: Optional[str]
+    token: Optional[str]
```
nikhilgarg28, line 355: Just to confirm, at what level do we do error handling in scenarios:
1. One of username or password is given but not both
2. Token is given along with one or both of username/password. 

It'd be good ensure that doing something like that creates an error with clear message in the mock world itself (ideally within __init__ function itself, not even commit).

nikhilgarg28: Assuming this is done (already or after some revisions), could you please add tests to catch these?

nikhilgarg28 (changes requested):
Requesting some changes

docs/pages/api-reference/connectors/kafka.md
saiharshavellanki: It doesn't work as of now. It needs a bit more work and tests. Post that we can add it to documentation

fennel/connectors/connectors.py
saiharshavellanki: Added validation and some tests

nikhilgarg28 (approved):

PR #508, opened 2024-07-10
[FIX] Added a more robust way to getting text code for where in preproc

nonibansal:
1. Moved removal of decorators from regex to AST.
2. Added more robust way to getting function text of a lambda.

This PR fixes most of the issues, but there are still some formatting for which this fails.

### DIFF
--- a/fennel/internal_lib/to_proto/source_code.py
+++ b/fennel/internal_lib/to_proto/source_code.py
@@ -5,10 +5,11 @@ import os
 import re
 import sys
 from textwrap import dedent, indent
-from typing import Callable, Dict, Optional
+from typing import Callable, Dict, Optional, List
 
 import fennel.gen.pycode_pb2 as pycode_proto
 from fennel.datasets import Dataset
+from fennel.datasets.datasets import get_all_operators
 from fennel.dtypes.dtypes import (
     FENNEL_STRUCT_SRC_CODE,
     get_fennel_struct,
@@ -18,7 +19,6 @@ from fennel.featuresets import Featureset
 from fennel.lib import FENNEL_GEN_CODE_MARKER
 from fennel.lib.includes.include_mod import FENNEL_INCLUDED_MOD
 from fennel.utils import fennel_get_source
-
 
 # Reserved column for server to do grouping windowing.
 SERVER_CLUSTER_COLUMN = "_@@_cluster"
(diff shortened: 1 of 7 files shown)


fennel/internal_lib/to_proto/source_code.py
```diff
@@ -101,15 +100,78 @@ def get_featureset_gen_code(
...
+                if isinstance(node, (ast.ClassDef, ast.FunctionDef))
+            ),
+            None,
+        )
+        for line in self.decorator_lines:
+            self.text = self.text.replace(line, "")
+        return self.text.strip()
+
+
+def remove_source_and_sink_decorator(text: str):
```
aditya-nambiar, line 173: Makes source and sink a parameter, so that it can be used by anyone else in the future. 

Can you add a test suite for this function, for different decorators say meta etc

fennel/internal_lib/to_proto/source_code.py
```diff
@@ -210,11 +282,96 @@ def lambda_to_python_regular_func(lambda_func):
...
+                if lambda_node.end_col_offset
+                else lambda_node.col_offset
+            ),
+        )
     )
+    lambda_next_line(lambda_func, lambda_text, line_num)
     return lambda_text
 
 
+def lambda_next_line(lambda_func: Callable, lambda_text: str, line_num: int):
```
aditya-nambiar, line 305: rename to the function to something like check_ambigous_lambda

fennel/internal_lib/to_proto/source_code.py
```diff
@@ -210,11 +282,96 @@ def lambda_to_python_regular_func(lambda_func):
...
+        lines = f.readlines()
+        if line_num < len(lines):
+            next_line = lines[line_num].strip()
+    # No Next line found
+    if next_line == "":
+        return
+    # Next line already in lambda_text
+    if next_line in lambda_text:
+        return
+    # If Next line is part on next operator
```
aditya-nambiar, line 322: this is just an optimization right ? I can totally see folks adding a new operator totally forgetting to add it here. 
One thing we could do is find all functions defined on _Node class and return early using that

fennel/internal_lib/to_proto/source_code.py
```diff
@@ -210,11 +282,96 @@ def lambda_to_python_regular_func(lambda_func):
...
+                        lambda_node.end_col_offset
+                        if lambda_node.end_col_offset
+                        else lambda_node.col_offset
+                    ),
+                )
+            )
+            if new_lambda_text != lambda_text:
+                raise ValueError(
+                    f"Not able to choose lambda function between {lambda_text} and {new_lambda_text}. "
+                    f"If lambda contains more than one line, please keep in one () bracket."
```
aditya-nambiar, line 369: please define the function within the parentheses

fennel/internal_lib/to_proto/source_code.py
```diff
@@ -101,15 +100,78 @@ def get_featureset_gen_code(
...
+    if lineno == end_lineno:
+        return lines[lineno - 1][col_offset:end_col_offset]
+    for i in range(lineno - 1, end_lineno):
+        if i == lineno - 1:
+            output.append(lines[i][col_offset:])
+        elif i == end_lineno - 1:
+            output.append(lines[i][:end_col_offset])
+        else:
+            output.append(lines[i])
+    return "\n".join(output)
```
aditya-nambiar, line 128: Can this be simplified - 

    # If the text is within a single line
    if lineno == end_lineno:
        return lines[lineno - 1][col_offset:end_col_offset]
    
    # Collect parts from multiple lines
    parts = [lines[lineno - 1][col_offset:]]  # Start from the initial column offset of the first line
    parts.extend(lines[lineno:end_lineno - 1])  # Add full lines between the first and last line
    parts.append(lines[end_lineno - 1][:end_col_offset])  # End at the final column offset of the last line    
    return "\n".join(parts)

fennel/internal_lib/to_proto/source_code.py
nonibansal: This was a needed thing, what was happening was something like event.filter(...) and if the next operator was coming in new line, that was getting appended to tha lambda function properly if the lambda expression was in a parenthesis. So lambda used to become like lambda x: (x["fdgd"] > 23).assign() or any filter operator and this to used to cause/return ambiguous lambda error. But yeah makes sense to fetch all the operators through some operation in _Node class.

fennel/internal_lib/to_proto/source_code.py
nonibansal: changed

fennel/internal_lib/to_proto/source_code.py
nonibansal: changed

fennel/internal_lib/to_proto/source_code.py
nonibansal: added.

fennel/internal_lib/to_proto/source_code.py
nonibansal: changed

aditya-nambiar (approved):

PR #516, opened 2024-07-19
Add more schema validations for select operator

nonibansal:


fennel/datasets/datasets.py
```diff
@@ -2770,12 +2796,46 @@ def visitDrop(self, obj) -> DSSchema:
...
+        key_fields = input_schema.keys
+
+        output_schema_name = f"'[Pipeline:{self.pipeline_name}]->select node'"
+        if obj.columns is None or len(obj.columns) == 0:
+            raise ValueError(
+                f"invalid select - {output_schema_name} must have at least one column to select"
+            )
+        if timestamp not in obj.columns:
+            raise ValueError(
+                f"invalid select - {output_schema_name} timestamp_field : `{timestamp}` must be in columns"
```
nikhilgarg28, line 2822: Super nit but can you please remove the underscore from `timestamp_field` and make it two words?

fennel/datasets/datasets.py
```diff
@@ -2770,12 +2796,46 @@ def visitDrop(self, obj) -> DSSchema:
...
+            )
+        for key_field in key_fields:
+            if key_field not in obj.columns:
+                raise ValueError(
+                    f"invalid select - {output_schema_name} key_field : `{key_field}` must be in columns"
```
nikhilgarg28, line 2827: Same here - key_field -> `key field`

nikhilgarg28 (approved):

fennel/datasets/datasets.py
nonibansal: changed

fennel/datasets/datasets.py
nonibansal: changed

PR #519, opened 2024-07-22
Increase timeout to 180 secs

nonibansal:


bothra90 (approved):

fennel/datasets/datasets.py
```diff
@@ -986,7 +986,8 @@ def __init__(
         self.node.out_edges.append(self)
 
     def signature(self):
-        return fhash(self.node.signature(), self.columns)
+        # Doing this to make signature same as previous one.
```
nikhilgarg28, line 989: Do we use signature anywhere on the backend now? If not, we should probably just get rid of it someday. also cc @aditya-nambiar if he has any opinions.

nikhilgarg28 (approved):

fennel/datasets/datasets.py
aditya-nambiar: yeah have been planning to get rid of it for a while

nonibansal: we use this in operator equality check just confirmed.

PR #522, opened 2024-08-01
Adding a param in checkout that allows init_branch

zerismo:
image


fennel/client/client.py
```diff
@@ -430,14 +430,17 @@ def list_branches(self) -> List[str]:
...
 
-    def checkout(self, name: str):
+    def checkout(self, name: str, init: bool = False):
         """Checkouts the client to another branch.
 
         Parameters:
         ----------
         name (str): The name of the branch to checkout.
 
         """
```
nikhilgarg28, line 440: Please also update the docstring here + docs.

docs/examples/api-reference/client/branch.py
```diff
@@ -55,4 +55,12 @@ def unused_checkout():
 
         # all subsequent operations will be on `mybranch`
         client.commit(...)
+
+        # create and change active branch from `mybranch` to `mybranch2`
+        # docsnip-highlight start
+        client.checkout("mybranch2", True)
```
nikhilgarg28, line 61: In code snippets, we should pass this init via kwarg instead of args. Basically whatever examples we write, people copy that and so we need our examples to set best practices on readability. 

`client.checkout("mybranch2", init=True)`

nikhilgarg28 (approved):

PR #523, opened 2024-08-02
Support for expressions

aditya-nambiar:


fennel/expr/test_expr.py
```diff
@@ -0,0 +1,316 @@
...
+    assert ret.tolist() == [9, 11, 13, 15]
+    ret = compute_expr(expr, df, TestDataset.schema())
+    assert ret.tolist() == [9, 11, 13, 15]
+    assert compile(expr, {"a": int, "b": int}) == int
+
+
+def test_math_expr():
+    expr = (F("a").num.floor() + 3.2).num.ceil()
+    printer = ExprPrinter()
+    expected = "CEIL((FLOOR(Field[`a`]) + 3.2))"
```
nikhilgarg28, line 69: Nit - let's print F('a') as Ref('a') vs Field. That way, it's going to be consistent with Rust backend + avoids confusion with dataset fields (since the same syntax needs to be reused for features later).

fennel/datasets/datasets.py
```diff
@@ -241,11 +245,50 @@ def transform(self, func: Callable, schema: Dict = {}) -> _Node:
...
+        1. Assigning a new column with a lambda function:
+            >>> ds.assign("new_column", int, lambda x: x.old_column + 1)
+        2. Using kwarg syntax to assign a column:
+            >>> ds.assign(new_column=F("old_column") + 1).astype(int))
+        3. Assigning multiple columns:
+            >>> ds.assign(
+            ...     new_column1=(F("old_column1") + 1).astype(int),
+            ...     new_column2=(F("old_column2") + 2).astype(int),
+            ... )
+
```
nikhilgarg28, line 263: 2 and 3 feel the same so I wonder if we can simplify the docstring a bit. Also, do we support the approach where it's 3 kwarg but they provide expression instead of lambda?

fennel/datasets/datasets.py
```diff
@@ -241,11 +245,50 @@ def transform(self, func: Callable, schema: Dict = {}) -> _Node:
...
+            and isinstance(args[2], Callable)
+        ):
+            return Assign(self, args[0], args[1], args[2])
+       
+        if len(kwargs) == 3:
+            # Check if the kwargs are name, dtype and func
+            if all(k in ["name", "dtype", "func"] for k in kwargs.keys()):
+                return Assign(
+                    self, kwargs["name"], kwargs["dtype"], kwargs["func"]
+                )
```
nikhilgarg28, line 283: Does this mean that when these magic kwargs are used, we can still fall to Pandas approach? Since this behavior is neither documented nor very ergonomic, I wonder if we should just remove it?

fennel/datasets/datasets.py
```diff
@@ -410,45 +453,100 @@ def dsschema(self):
...
+
 class Assign(_Node):
     def __init__(
-        self, node: _Node, column: str, output_type: Type, func: Callable
+        self,
+        node: _Node,
+        column: Optional[str],
+        output_type: Optional[Type],
+        func: Optional[Callable],
+        **kwargs,
```
nikhilgarg28, line 468: Maybe do some validation here that if kwargs are given, all of column, output_type, func are None. And if any of them is non-None, all are non-None and kwargs is empty.

fennel/expr/expr.py
```diff
@@ -0,0 +1,748 @@
...
+        self.out_edges = []
+        self.root = self if root is None else root
+        self.dtype = dtype
+
+    def edge(self, node):
+        self.out_edges.append(node)
+
+    def astype(self, dtype: Type) -> Expr:
+        self.dtype = dtype
+        return self
```
nikhilgarg28, line 28: JFYI - the way this is written, someone could put `astype` in the middle of the expression or put `astype` twice. You may want to consider an alternative where there are two classes `Expr` and `TypedExpr` or something like that. TypedExpr has no other methods so is terminal. astype on Expr takes you to TypedExpr.

fennel/expr/expr.py
```diff
@@ -0,0 +1,748 @@
...
+            return f"{self.left}[{self.right}]"
+        else:
+            return f"{self.left} {self.op} {self.right}"
+
+
+class When(Expr):
+    def __init__(self, expr: Expr, root: Optional[Expr] = None):
+        self.expr = make_expr(expr)
+        self.expr.edge(self)
+        print("Created when")
```
nikhilgarg28, line 642: Remove all stray prints, at least in non-test code.

fennel/expr/expr.py
```diff
@@ -0,0 +1,748 @@
...
+
+class InvalidExprException(Exception):
+    pass
+
+
+class Expr(object):
+    def __init__(self, root=None, dtype=None):
+        self.nodeid = id(self)
+        self.inline = False
+        self.out_edges = []
```
nikhilgarg28, line 19: What is this for?

fennel/expr/expr.py
```diff
@@ -0,0 +1,748 @@
...
+
+    def __rxor__(self, other: Any):
+        raise InvalidExprException("binary operation 'xor' not supported")
+
+    def varname(self, dollar: bool = True) -> str:
+        name = self.nodeid
+        if dollar:
+            return f"${name}"
+        else:
+            return f"{name}"
```
nikhilgarg28, line 301: copied from rexer? :) probably not used here though, right?

fennel/expr/expr.py
```diff
@@ -0,0 +1,748 @@
...
+        super(Binary, self).__init__(None, dtype)
+
+    def __str__(self) -> str:
+        if self.op == "[]":
+            return f"{self.left}[{self.right}]"
+        else:
+            return f"{self.left} {self.op} {self.right}"
+
+
+class When(Expr):
```
nikhilgarg28, line 638: When is an expresssion so technically one could write `when(something).list(...)` right? do we have protections in place to ensure that when must be followed by then?

fennel/expr/test_expr.py
```diff
@@ -0,0 +1,316 @@
...
+from typing import Dict
+from fennel.datasets import dataset
+
+from fennel.expr import F, when
+from fennel.expr.visitor import ExprPrinter
+from fennel.expr.serializer import ExprSerializer
+from google.protobuf.json_format import ParseDict  # type: ignore
+from fennel.gen.expr_pb2 import Expr
+from fennel.testing.test_utils import error_message
+from fennel.expr.utils import compute_expr, compile
```
nikhilgarg28, line 13: rename compute_expr -> `eval` and compile -> `typeof`

nikhilgarg28: You could also consider adding them as methods on `Expr` itself instead of adding them to utils.

fennel/datasets/test_invalid_dataset.py
```diff
@@ -220,6 +221,77 @@ class RatingActivity:
...
+            def transform(cls, rating: Dataset):
+                return rating.assign(
+                    rating_sq=(F("rating") * F("rating")).astype(str),
+                    movie_suffixed=F("movie").str.concat("_suffix").astype(int),
+                ).drop("rating","movie")
+
+    assert (
+        str(e.value)
+        == """invalid assign - '[Pipeline:transform]->assign node' type errors: [TypeError("expression `(Field[`rating`] * Field[`rating`])` has type `<class 'float'>` does not match the type of column `rating_sq`, <class 'str'>"), TypeError('expression `Field[`movie`] + "_suffix"` has type `<class \\\'str\\\'>` does not match the type of column `movie_suffixed`, <class \\\'int\\\'>')]"""
+    )
```
nikhilgarg28, line 246: I think we should put it some work to significantly polish this error because it's going to be a common occurrence. Here is my attempt at what it should be:

`type error:  '{kwarg}' in assign operator of '{DS}.{pipeline}' is of type float, can not be promoted to str. Full expression: {expression}`

fennel/datasets/test_invalid_dataset.py
```diff
@@ -220,6 +221,77 @@ class RatingActivity:
...
+            def transform(cls, rating: Dataset):
+                return rating.assign(
+                    rating_sq=(F("rating") * F("rating")).astype(float),
+                    movie_suffixed=F("movie").str.concat("_suffix").astype(str),
+                ).drop("rating","movie")
+
+    assert (
+        str(e2.value)
+        == """[TypeError('Field `rating_sq` has type `float` in `pipeline transform output value` schema but type `int` in `RatingActivityTransformed2 value` schema.')]"""
+    )
```
nikhilgarg28, line 269: This error is also very hard to read. Suggest to simplify this as P0.

fennel/datasets/test_invalid_dataset.py
```diff
@@ -220,6 +221,77 @@ class RatingActivity:
...
+            @inputs(RatingActivity)
+            def transform(cls, rating: Dataset):
+                return rating.assign(
+                    rating_sq=(F("rating") % F("rating")).astype(float),
+                    movie_suffixed=(F("movie") + "_suffix").astype(str),
+                ).drop("rating","movie")
+
+    assert (
+        str(e2.value)
+        == """invalid assign - '[Pipeline:transform]->assign node' error in expression for column `movie_suffixed`: Failed to compile expression: invalid expression: both sides of '+' must be numeric types but found String & String, left: col(movie), right: lit(String("_suffix"))"""
```
nikhilgarg28, line 291: I reckon this error is coming from Rust lib? If so, would strongly suggest to make sure that both Rust/Python print the same expression. For instance, this one uses `col` (with round brackets) and `String` but Python version uses `Field` with square brackets and `str`. Can we not use the print from the rust library itself in Python so there is a single copy of this?

fennel/datasets/test_invalid_dataset.py
```diff
@@ -220,6 +221,77 @@ class RatingActivity:
     t: datetime
 
 
+def test_incorrect_assign_expr_type():
```
nikhilgarg28, line 224: Could you plz add one test case for filter as well where the expression doesn't evaluate to bool?

fennel/datasets/datasets.py
aditya-nambiar: Yeah the former is args and if they do indeed use kwarg we typecheck. But looks like you suggested to remove the latter

fennel/datasets/test_invalid_dataset.py
aditya-nambiar: Here i am combining all errors that belong to the same pipeline so will make it - 

type error(s) in assign operator of '{DS}.{pipeline}': 
1. '{kwarg}'  is of type float, can not be promoted to str. Full expression: {expression}
2. '{kwarg2}'  is of type int, can not be promoted to str. Full expression: {expression}

fennel/datasets/datasets.py
```diff
@@ -410,44 +453,113 @@ def dsschema(self):
...
+            self.assign_type = UDFType.expr
+
+    @classmethod
+    def from_expressions(cls, self, **kwargs):
+        for k, v in kwargs.items():
+            if not isinstance(v, TypedExpr):
+                raise ValueError(
+                    "Assign.from_expressions expects all values to be of type Expr",
+                    f"found `{type(v)}` for column `{k}`",
+                )
```
nikhilgarg28, line 506: Nit but you might want to add a special check where it's not `TypedExpr` but is `Expr` and give an error message that is more specific.

fennel/expr/test_expr.py
```diff
@@ -0,0 +1,314 @@
...
+from google.protobuf.json_format import ParseDict  # type: ignore
+from fennel.gen.expr_pb2 import Expr
+from fennel.testing.test_utils import error_message
+
+
+def test_basic_expr1():
+    expr = (F("num") + F("d")).isnull()
+    df = pd.DataFrame({"num": [1, 2, 3, 4], "d": [5, 6, 7, 8]})
+    assert expr.typeof({"num": int, "d": int}) == bool
+    ret = expr.eval(df, {"num": int, "d": int})
```
nikhilgarg28, line 19: Nit but this isn't looking very readable. I think it might be better to make `schema` a kwargs. It may end up just being about writing examples/docs where we express it like kwargs vs args without any change to the `typeof` or `eval` functions.

fennel/internal_lib/schema/test_schema.py
```diff
@@ -799,3 +803,158 @@ def test_convert_dtype_to_arrow_type():
...
     )
+
+
+class TestDataTypeConversions(unittest.TestCase):
+
+    def test_int_type(self):
+        original = int
+        proto = get_datatype(original)
+        result = from_proto(proto)
+        self.assertEqual(original, result)
```
nikhilgarg28, line 814: Super nit but this would be better written as case builder pattern. Where we have a list of lots of cases and for each case, we make a simple check in a for loop. Makes it easy to maintain/add more tests with low overhead.

nikhilgarg28 (approved):

fennel/expr/test_expr.py
aditya-nambiar: yeah we should do that in the examples

PR #526, opened 2024-08-06
extractor: Support chained extractors

aditya-nambiar:


fennel/featuresets/featureset.py
```diff
@@ -72,13 +72,19 @@ def feature(
     **kwargs,
 ) -> T:  # type: ignore
     if len(args) > 1:
-        raise ValueError(
+        raise TypeError(
             f"We can only reference to only one feature or one field at a time found : {args}"
```
nikhilgarg28, line 76: Unrelated but this error looks ugly. We don't use 3rd person (i.e. "we") in error messages anywhere else AFAIK.

nikhilgarg28 (approved):

PR #531, opened 2024-08-19
expr: Read side expressions

aditya-nambiar:


docs/examples/api-reference/client.py
```diff
@@ -41,7 +41,7 @@ class UserFeatures:
     name: str
     country_geoid: int
     # The users gender among male/female/non-binary
-    age: int = F().meta(owner="aditya@fennel.ai")
+    age: int = feature().meta(owner="aditya@fennel.ai")
```
nikhilgarg28, line 44: I think there is a place in docs where we suggest this as a common shorthand to use - could you plz change/remove that too? https://fennel.ai/docs/concepts/featureset#convention

fennel/expr/test_expr.py
```diff
@@ -94,17 +101,31 @@ def test_math_expr():
...
+            "d": [1, 2, 3, 4],
+        }
     )
-    df = pd.DataFrame({"a": [1.4, 3.2, 6.1, 4.8], "b": [100, 200, 300, 400]})
-    ret = expr.eval(df, {"a": float, "b": int})
-    assert ret.tolist() == [1, 3.2, 300, 4.8]
-    assert expr.typeof({"a": float, "b": int}) == float
+    ret = expr.eval(df, {"a": float, "b": int, "d": int})
+    assert ret.tolist() == [2.0, 3.2, 300, 4.8]
+    assert expr.typeof({"a": float, "b": int, "d": int}) == float
```
nikhilgarg28, line 124: Unrelated to this PR but since `typeof` only takes schema, why not pass this as kwargs instead of a dictionary? Like

`expr.typeof(a=float, b=int, d=int)`

Don't feel very strongly though, upto you. Maybe we can support both?

aditya-nambiar: So that we can pass dataset.schema() to it directly

fennel/featuresets/featureset.py
```diff
@@ -551,6 +562,70 @@ def _add_feature_names_as_attributes(self):
...
+            )
+            # Converting the string in extractor inputs decorator to Feature object
+            inputs = []
+            for input in extractor.user_defined_inputs:
+                if isinstance(input, str):
+                    try:
+                        inputs.append(self._feature_map[input])
+                    except KeyError:
+                        raise ValueError(
+                            f"When using F(<feature_name>) for expressions, one can only choose from the features defined in the current featureset. "
```
nikhilgarg28, line 598: FYI - this error will change when we change `F` to `col`

nikhilgarg28: The language also need a bit work:

`expression refers to feature col("{}") not present in the featureset {}`

fennel/featuresets/featureset.py
```diff
@@ -551,6 +562,70 @@ def _add_feature_names_as_attributes(self):
...
+                            f"Please choose an input from : {list(self._feature_map.keys())} found : `{input}` in extractor : `{extractor.name}`."
```
nikhilgarg28, line 599: Remove all this?

fennel/featuresets/featureset.py
```diff
@@ -551,6 +562,70 @@ def _add_feature_names_as_attributes(self):
...
+                        )
+                elif isinstance(input, Feature):
+                    inputs.append(input)
+                else:
+                    raise ValueError(
+                        f"Parameter `{input}` is not a feature, but a "
+                        f"`{type(input)}`, and hence not supported as an input for the extractor "
+                        f"`{extractor.name}`"
```
nikhilgarg28, line 607: `extractor for '{}' refers to {input} of type {type(input)}, expected str`

fennel/featuresets/featureset.py
```diff
@@ -551,6 +562,70 @@ def _add_feature_names_as_attributes(self):
...
+                    )
+
+            extractor.expr = expr
+            extractor.inputs = inputs
+            input_types = {inp.name: inp.dtype for inp in inputs}
+            computed_dtype = expr.typeof(input_types)
+            if computed_dtype != feature.dtype:
```
nikhilgarg28, line 614: We don't need `typeof` to be same as feature.dtype, only that computed_dtype can be promoted into feature.dtype. So if it computes int but the type is float/option<int> etc, it's fine.

nikhilgarg28: The matches function on compiled expression does this for you already.

fennel/featuresets/featureset.py
```diff
@@ -551,6 +562,70 @@ def _add_feature_names_as_attributes(self):
...
+                raise TypeError(
+                    f"Expected expression {expr} for feature `{feature.name}` to return dtype `{dtype_to_string(feature.dtype)}` "
+                    f"but got `{dtype_to_string(computed_dtype)}`"
```
nikhilgarg28, line 617: `expression {expr} for feature {feature.name} is of type `dtype_to_string(feature.dtype)` but got {other_type}`

fennel/featuresets/featureset.py
```diff
@@ -74,13 +77,17 @@ def feature(
...
         )
     if default is not None:
         if len(args) == 0:
             raise TypeError(
                 'Please specify a reference to a field of a dataset to use "default" param'
             )
+        if isinstance(args[0], Expr):
+            raise ValueError(
+                f"Default value cannot be set for an expression, use fillnull instead, found : {args[0]}"
+            )
```
nikhilgarg28, line 90: `error in extractor of {feature}; can not set default value for expressions, maybe use fillnull instead?`

fennel/featuresets/test_invalid_featureset.py
```diff
@@ -377,11 +378,81 @@ class A:
...
+            user_id: int
+            home_geoid: int
+            age: int = feature(UserInfoDataset.age, default=0)
+            age_squared: str = feature(F("age") * F("age"))
+            credit_score: int
+
+    assert (
+        str(e.value)
+        == "Expected expression Ref('age') * Ref('age') for feature `age_squared` to return dtype `str` but got `int`"
+    )
```
nikhilgarg28, line 440: This shouldn't say `Ref` but `col`, same as Python syntax. I guess it's a Rust side change? If so, I can fix it as part of my PR.

fennel/testing/query_engine.py
```diff
@@ -346,6 +353,30 @@ def _check_schema_exceptions(
...
 
+    def _compute_expr_extractor(
+        self,
+        extractor: Extractor,
+        intermediate_data: Dict[str, pd.Series],
+    ) -> pd.Series:
+        if len(extractor.outputs) != 1:
+            raise ValueError(
+                f"Expr extractor {extractor.name} must have exactly one output feature, found {len(extractor.outputs)}"
+            )
```
nikhilgarg28, line 364: `extractor for feature {} outputs {len(extractor.outputs)} features, expected one`

fennel/testing/query_engine.py
```diff
@@ -346,6 +353,30 @@ def _check_schema_exceptions(
...
+        if len(extractor.depends_on) != 0:
+            raise ValueError(
+                f"Expr extractor {extractor.name} must have exactly zero dependent datasets, found {len(extractor.depends_on)}"
+            )
```
nikhilgarg28, line 368: `extractor for feature {} depends on {len(dataset)} datasets, expression based extractors can not depend on datasets`

nikhilgarg28 (approved):
Mostly leaving comments on improving error messages. As a meta thought, let's refrain from printing names of auto generated extractors in error messages. Only correctness issue is around type promotion check at commit time.

Rest everything is good.

PR #535, opened 2024-08-21
Add fields parameter in join to fetch a subset of fields from rhs

satrana42:
Add right_fields parameter in join to fetch only a subset of fields from rhs


fennel/client_tests/test_dataset.py
```diff
@@ -1420,6 +1420,126 @@ def pipeline_join(cls, info: Dataset, sale: Dataset):
...
+            return df[
+                [
+                    str(cls.movie),
+                    str(cls.t),
+                    str(cls.revenue_in_millions),
+                    str(cls.rating),
+                ]
+            ]
+
+        c = rating.join(revenue, how="left", on=[str(cls.movie)], right_fields=["revenue"])
```
aditya-nambiar, line 1460: s/right_fields/rhs_fields

fennel/datasets/test_invalid_dataset.py
```diff
@@ -768,6 +768,96 @@ def create_pipeline(cls, a: Dataset, b: Dataset):
...
+            age: int
+            timestamp: datetime
+
+            @pipeline
+            @inputs(XYZ, ABC)
+            def create_pipeline(cls, a: Dataset, b: Dataset):
+                c = a.join(b, how="inner", on=["user_id"], right_fields=["rank"])  # type: ignore
+                return c
+
+    assert "doesn't exist in allowed fields of right schema" in str(e.value)
```
aditya-nambiar, line 799: can we add assert  that we print the pipeline and dataset name in the error message so that user can figure where the issue is

fennel/datasets/test_invalid_dataset.py
```diff
@@ -768,6 +768,96 @@ def create_pipeline(cls, a: Dataset, b: Dataset):
...
+            age: int
+            timestamp: datetime
+
+            @pipeline
+            @inputs(XYZ, ABC)
+            def create_pipeline(cls, a: Dataset, b: Dataset):
+                c = a.join(b, how="inner", on=["user_id"], right_fields=["user_id"])  # type: ignore
+                return c
+
+    assert "doesn't exist in allowed fields of right schema" in str(e.value)
```
aditya-nambiar, line 828: user doesnt know what the allowed fields are? can we tell them eg value fields and or timestamp of the rhs dataset

fennel/datasets/test_invalid_dataset.py
```diff
@@ -768,6 +768,96 @@ def create_pipeline(cls, a: Dataset, b: Dataset):
...
+            age: int
+            timestamp: datetime
+
+            @pipeline
+            @inputs(XYZ, ABC)
+            def create_pipeline(cls, a: Dataset, b: Dataset):
+                c = a.join(b, how="inner", on=["user_id"], right_fields=["timestamp"])  # type: ignore
+                return c
+
+    assert "already exists in left schema" in str(e.value)
```
aditya-nambiar, line 857: please assert on teh full message rather than using "in"

fennel/datasets/test_dataset.py
```diff
@@ -1259,20 +1264,21 @@ def pipeline1(cls, a: Dataset, b: Dataset):
...
+	    "datasetName":"ABDataset",
+	    "join":{
+	        "lhsOperandId":"A",
+	        "rhsDsrefOperandId":"B",
+	        "on":{
+	            "a1":"b1"
+	        },
+	        "withinLow":"259200s",
+	        "withinHigh":"31536000s"
+	    },
```
aditya-nambiar, line 1279: why is the how missing ?  We should not modify what the protos used to generate since that is stored state on disk

fennel/datasets/datasets.py
```diff
@@ -1030,15 +1037,32 @@ def make_types_optional(types: Dict[str, Type]) -> Dict[str, Type]:
         if self.how == "left":
             right_value_schema = make_types_optional(right_value_schema)
 
-        # Add right value columns to left schema. Check for column name collisions
+        # TODO(sat): Are the same checks here and in SchemaValidator redundant?
+        # If right_fields is set, check that it contains elements from right schema values and timestamp only
+        if self.right_fields is not None and len(self.right_fields) > 0:
```
aditya-nambiar, line 1042: prefer to raise error in SchemaValidator as we can give a lot more context about the error there

fennel/client_tests/test_dataset.py
satrana42: I used right_fields because the param for join on is called right_on and not rhs_on.

nikhilgarg28: I wonder if we should just use `fields` here since that's quite disambiguated on its own (there is no analog of left_fields)? cc @aditya-nambiar

fennel/datasets/test_dataset.py
satrana42: Shouldn't the test throw an error if I copied wrong json?

aditya-nambiar (approved):

PR #538, opened 2024-08-23
Add assertions for 1:1 mapping of rows for extractors

zerismo:


fennel/client_tests/test_featureset.py
```diff
@@ -239,6 +258,33 @@ def test_simple_extractor(self, client):
...
+        client.sleep()
+        ts = pd.Series([now, now])
+        user_ids = pd.Series([18232, 18234])
+        age = pd.Series([18, 22])
+        with pytest.raises(AssertionError) as exc_info:
+            UserInfoSingleErrorProneExtractor.get_user_info(
+                UserInfoMultipleExtractor, ts, user_ids, age
+            )
+        assert "Output length mismatch" in str(exc_info.value)
+
```
aditya-nambiar, line 287: can we also list out the differences in length

fennel/featuresets/featureset.py
```diff
@@ -450,6 +450,63 @@ def is_user_defined(obj):
...
+        input_lengths = [len(series) for series in input_series]
+
+        # Check if all input series have the same length
+        first_length = input_lengths[0]
+        
+        for i, series in enumerate(input_series):
+            assert len(series) == first_length, (
+                f"Input length mismatch: expected {first_length}, got {len(series)} for input {i+1}"
+            )
+
```
aditya-nambiar, line 483: this is provided by us and you can assume that is correct

PR #540, opened 2024-08-27
expr: Make it hashable

aditya-nambiar:


nonibansal (approved):

fennel/utils.py
```diff
@@ -163,6 +163,12 @@ def _json_default(item: Any):
     if isinstance(item, datetime.timedelta):
         return str(item.total_seconds())
 
+    if isinstance(item, TypedExpr):
+        return str(item.expr) + str(item.dtype)
```
nikhilgarg28, line 167: Better to return these as tuples instead of concact as strs for the purpose of hash. That way, you may have cases where the concatenated string ends up being another valid string of another type. Not possible at all with tuples since they become nested.

aditya-nambiar: you cant create a string from a type and anyways these signatures are pointless, need to remove them from the codebase when i get more time

PR #545, opened 2024-08-29
Support S3 batch sink

saiharshavellanki:
- Add support for S3 batch sink
- Updated the documentation accordingly
Screenshot 2024-08-29 at 11 39 52 AM

Screenshot 2024-08-29 at 11 40 09 AM


nikhilgarg28:
Nit on docs - in a couple of places (e.g. style), you are using 'We'. Could you please make all that 3rd person instead? So 'we do x...' or 'we don't allow x...' -> "Fennel does X..." or "Fennel doesn't allow you to ..."

nikhilgarg28:
Also, don't we need to expose 'format' here?

saiharshavellanki:
> Also, don't we need to expose 'format' here?

Format is exposed as part of ExtTable for S3. I am following the semantics similar to Sources

nikhilgarg28 (approved):

PR #547, opened 2024-08-30
Assign in preproc

nonibansal:


nikhilgarg28 (approved):
Looks good. Though request to add one special test case - 
1. Incoming field `f` is of type `Optional[int]` (but in the dataset, it will be a key field of type `int`)
2. We use eval to make `f` -1 when it's None (but still store it in field of name `f`)
3. There are some other preproc ref/const in the same source for other fields.
4. We also have a `where` filter in source that only keeps the row when `f` is not -1.

nonibansal:
> Looks good. Though request to add one special test case -
> 
> 1. Incoming field `f` is of type `Optional[int]` (but in the dataset, it will be a key field of type `int`)
> 2. We use eval to make `f` -1 when it's None (but still store it in field of name `f`)
> 3. There are some other preproc ref/const in the same source for other fields.
> 4. We also have a `where` filter in source that only keeps the row when `f` is not -1.

added

PR #555, opened 2024-09-08
added support for secrets manager

hemanthk269:
- Added support for sending secret id and secret mapping from client to server
- User can add secret_id and mapping instead of username and password as follows

```
postgres = Postgres(
    name="postgres",
    host="localhost",
    db_name="test",
    username="username",
    password="password",
)

postgres_with_secret_id = Postgres(
    name="postgres",
    host="localhost",
    db_name="test",
    username_password_secret_id="secret_id",
    username_password_secret_mapping={"username": "<username-key>", "password": "<password-key>"},
)
```

Added validations like 
1. either username and password or username_password_secret_id and mapping should be there
2. only username and password key should be present in the mapping (not needed for bigquery and pub sub)


hemanthk269:
implementing in https://github.com/fennel-ai/client/pull/557

PR #558, opened 2024-09-11
expr: Use expr root in extractors

aditya-nambiar:


nonibansal (approved):

fennel/internal_lib/to_proto/source_code.py
```diff
@@ -414,7 +414,7 @@ def get_all_imports() -> str:
         "from fennel.lib.metadata import meta",
         "from fennel.lib import secrets, bucketize",
         "from fennel.datasets.datasets import dataset_lookup",
-        "from fennel.expr import col",
+        "from fennel.expr import col, lit, when",
```
nonibansal, line 417: @aditya-nambiar let's also add common functions from expr to imports like make_struct, etc.

aditya-nambiar: i'l add that later, since make_struct is new and if any tier doesn't have the latest client it will break

PR #560, opened 2024-09-13
Support date and datetime types in Min and Max aggregations

satrana42:
Min and Max aggregations support finding minimum or maximum of integer, float and
decimal data types. This change adds support for date and datetime types.


satrana42:
This is WIP, unit test passes but integration test fails.

aditya-nambiar (approved):
Do increment python version and add to changelog before landing

PR #563, opened 2024-09-17
lambda: Fix bug for using lambdas in jupyter notebook

aditya-nambiar:


fennel/internal_lib/to_proto/source_code.py
```diff
@@ -321,11 +321,18 @@ def check_ambiguous_lambda(
...
-        lines = f.readlines()
-        if line_num < len(lines):
-            next_line = lines[line_num].strip()
-
+    try:
+        with open(lambda_func.__code__.co_filename, "r") as f:
+            lines = f.readlines()
+            if line_num < len(lines):
+                next_line = lines[line_num].strip()
+    except Exception:
```
bothra90, line 329: Can you add a comment for when we expect the exception to happen what we're trying to do? Is the behavior desirable in general or only if we're operating in a notebook environment? If the latter, can we somehow check for that if this is not always the right thing to do?

bothra90 (approved):

PR #578, opened 2024-10-07
Add changelog operator

satrana42:
In this change we add changelog operator which converts a deltaframe into an append only
changelog, with keys and kinds stored in value columns.


fennel/client_tests/test_dataset.py
```diff
@@ -4658,30 +4658,50 @@ def rank_job(cls, job_opening: Dataset):
...
     openings = pd.DataFrame(
         {
             "creater_id": [1, 2, 1, 1, 1, 1],
             "job_id": [1, 2, 3, 4, 5, 6],
             "creation_ts": creation_ts,
         }
     )
-    log(JobOpening, openings)
+    client.log("fennel_webhook", "JobOpening", openings)
+
```
aditya-nambiar, line 4670: Making kind a key is a wierd thing tbh, I was expecting creater_id and maybe join_id to be the key.
In one test you do 

.latest().changelog() and assert that the chanelog produce the correct inserts and delete ( assert against client.get_dataset_df )

and another test

where we take dataset A

then derive B from A which does .groupby.latest().changelog.filter(filter deletes)groupby.latest() and we should get A back

fennel/testing/executor.py
```diff
@@ -954,8 +954,8 @@ def visitChangelog(self, obj):
         delete_df[input_ret.timestamp_field] = delete_df[
             FENNEL_DELETE_TIMESTAMP
         ]
-        df[obj.delete_column] = False
-        delete_df[obj.delete_column] = True
+        df[obj.kind_col] = "insert"
+        delete_df[obj.kind_col] = "delete"
```
aditya-nambiar, line 958: Lets convert these to bools

aditya-nambiar (changes requested):

fennel/client_tests/test_dataset.py
satrana42: I made `kind` a key because -
a) `get_datatset_df` is not defined for integration client, so couldn't test on the whole output of `changelog`
b) `latest` only picks one row for a timestamp randomly (atleast in server code), and we have inserts and deletes at the same timestamp.

aditya-nambiar: >latest only picks one row for a timestamp randomly (atleast in server code), and we have inserts and deletes at the same timestamp.
that is expected behavior, when you look up you should see the insert.

Anycase, i guess this test gets the job done, so marking it resolved

aditya-nambiar (approved):

PR #586, opened 2024-10-18
Client Join Improvement

petrpan26:
Two improvements was made here:

+ Join time is now from LHS to match server logic. Table join we might need to do differently but @bothra90 is working on that
+ Join now doesn't require index anymore. This is still blocked by server but will be addressed soon

----

> [!IMPORTANT]
> Align join time with server logic and remove right index requirement in join functionality.
> 
>   - **Behavior**:
>     - Align join time logic with server by using LHS timestamp in `executor.py`.
>     - Remove right index requirement in `datasets.py`.
>   - **Testing**:
>     - Remove `test_join_with_wrong_right_index()` from `test_schema_validator.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=fennel-ai%2Fclient&utm_source=github&utm_medium=referral)<sup> for 7c8ec095f6ec51c9f184650903f9c3edc99e9833. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/fennel/datasets/test_schema_validator.py
+++ b/fennel/datasets/test_schema_validator.py
@@ -1510,42 +1510,6 @@ def test_window_invalid_gap():
     )
 
 
-def test_join_with_wrong_right_index():
-    with pytest.raises(ValueError) as e:
-
-        @meta(owner="nitin@fennel.ai")
-        @dataset(online=True)
-        class Users:
-            user_id: str = field(key=True)
-            age: int
-            t: datetime
-
-        @meta(owner="nitin@fennel.ai")
-        @dataset
-        class Login:
-            user_id: str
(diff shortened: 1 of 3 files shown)


fennel/datasets/test_schema_validator.py
```diff
@@ -1510,42 +1510,6 @@ def pipeline_window(cls, app_event: Dataset):
     )
 
 
-def test_join_with_wrong_right_index():
```
nikhilgarg28, line 1513: Instead of removing it, can you instead modify it to confirm that it's possible to join with a dataset without the right index?

nikhilgarg28 (approved):
Approving tentatively.

PR #638, opened 2025-02-13
Add num dpus parameter to query offline API

saiharshavellanki:
Add `num_dpus` parameter to `query_offline()` in `client.py` and `mock_client.py` to specify Glue DPUs for offline queries.

### DIFF
--- a/fennel/CHANGELOG.md
+++ b/fennel/CHANGELOG.md
@@ -1,4 +1,6 @@
 # Changelog
+## [1.5.53] - 2024-11-12
+- Add support for num_dpus parameter in offline query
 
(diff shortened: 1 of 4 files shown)


fennel/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Changelog
+## [1.5.53] - 2024-11-12
```
nonibansal, line 2: change 53 to correct number

nonibansal (changes requested):
Upgrade version in pyproject.toml file and update documentation.

nonibansal (approved):

